Issue #5, opened 2026-04-05
vault_read / vault_search / vault_batch_read fail with "Object of type date is not JSON serializable"


mleitnercom:
## Description

All read and search operations fail when a vault file contains YAML frontmatter with bare (unquoted) date values.

**Affected tools:** `vault_read`, `vault_search`, `vault_batch_read`
**Not affected:** `vault_list` (does not parse frontmatter content)

## Steps to Reproduce

1. Create a markdown file with a bare date in frontmatter:

```yaml
---
created: 2026-04-05
tags: [test]
---
# Test Note
Some content.
```

2. Call `vault_read` on this file via the MCP.

3. **Expected:** File content and metadata returned as JSON.
4. **Actual:** Error: `Object of type date is not JSON serializable`

## Root Cause

Python's YAML parser (`PyYAML` / `ruamel.yaml`) automatically converts bare ISO date strings (e.g., `2026-04-05`) into `datetime.date` objects. When the parsed frontmatter is later serialized to JSON via `json.dumps()`, the serialization fails because `json.dumps` does not handle `datetime.date` natively.

Quoting the date (`created: "2026-04-05"`) causes YAML to parse it as a string, which works — but this is a workaround, not a fix. Users should not need to quote their dates for the MCP to function.

## Suggested Fix

Add a custom JSON encoder or `default` handler that converts `datetime.date` and `datetime.datetime` objects to ISO format strings:

```python
import json
from datetime import date, datetime

class DateAwareEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, (date, datetime)):
            return obj.isoformat()
        return super().default(obj)

# Usage:
json.dumps(data, cls=DateAwareEncoder)
```
Alternatively, a simpler approach with the `default` parameter:

```python
json.dumps(data, default=lambda obj: obj.isoformat() if isinstance(obj, (date, datetime)) else str(obj))
```

## Impact

This bug makes the MCP unusable for any vault where files use standard Obsidian frontmatter with bare date fields — which is the default behavior when using Obsidian's Properties panel or templates. Only `vault_list` works; all content-reading operations fail.

## Workaround

Quote all date values in frontmatter:

```yaml
# Before (fails):
created: 2026-04-05

# After (works):
created: "2026-04-05"
```

jjsmackay:
[...]

PR #7, opened 2026-04-11
Fix MCP spec 2025-06-18 compliance, OAuth persistence, and session-lifecycle bugs

marcelotoledo:
## Summary

This PR bundles five independent fixes found while deploying the server behind a Tailscale tunnel and connecting it to claude.ai. Each one is a standalone bug; happy to split into separate PRs if preferred.

### 1. Constant-time token comparison (`auth.py`)
`BearerAuthMiddleware` compared the bearer token with `!=`, which is timing-variable. Switched to `hmac.compare_digest()`.

### 2. `/authorize` alias for OAuth discovery (`oauth.py`, `auth.py`)
claude.ai's MCP connector calls `/authorize` directly (without the `/oauth/` prefix) in addition to what the discovery document advertises. The server returned 401 because (a) no route existed at `/authorize` and (b) even if it did, the bearer middleware would block it. Added `/authorize` as a route alias to the same handler and added it to `_AUTH_EXEMPT_PATHS`.

### 3. MCP spec 2025-06-18 compliance (`server.py`, `auth.py`)
Three related changes needed to match what the current spec and claude.ai expect:
- Transport now served at `/` instead of `/mcp/` (via `streamable_http_path="/"`).
- Added a `HEAD /` and `GET /` probe handler returning `MCP-Protocol-Version: 2025-06-18`. To keep this public without opening the POST surface, added a method-aware exempt set `_AUTH_EXEMPT_METHOD_PATHS = {("GET", "/"), ("HEAD", "/")}` — POST `/` continues to require the bearer token.
- Switched `stateless_http=True` → `False` so FastMCP emits and validates `Mcp-Session-Id` per the spec.

### 4. Persist dynamically-registered OAuth clients (`oauth.py`, `.gitignore`)
`oauth_register` generated a client_id and returned the env-var secret, but never stored anything. Every restart effectively invalidated any client that had used dynamic registration. Now each registration gets a unique `client_secret` and is persisted to `.oauth_clients.json` at the project root (atomic write via tempfile + `os.replace`, `chmod 600`, loaded on module import). `_handle_client_credentials` checks the registered store first and falls back to the env-var client so the README flow still works. `.oauth_clients.json` added to `.gitignore`.

### 5. Move vault observer to process lifecycle (`server.py`, `frontmatter_index.py`)
With `stateless_http=False`, FastMCP runs the lifespan per MCP session. Every new `initialize` called `frontmatter_index.start()` again on the same `Observer`, which crashed on macOS with `RuntimeError: Cannot add watch - it is already scheduled` from `FSEventsEmitter`. The index is a process-wide singleton (tools already import it directly from the module), so its lifecycle belongs to the process, not the session:
- `start()` moved into `main()`, once, before `uvicorn.run()`.
- `stop()` registered via `atexit`.
- `lifespan()` no longer touches start/stop; it only yields the shared instance.
- Defensive `if self._observer is not None: return` guard at the top of `start()` makes double-calls a no-op regardless of call site.

## Test plan
- [ ] `curl -I http://localhost:8420/` returns `200` with `MCP-Protocol-Version: 2025-06-18` and no auth required.
- [ ] `curl -X POST http://localhost:8420/` without bearer returns `401`.
- [ ] OAuth authorization code flow via `/authorize` completes end-to-end.
- [ ] Connect via claude.ai, restart the server, verify the integration reconnects without re-registering (check that `.oauth_clients.json` is populated).
- [ ] Reconnect claude.ai multiple times in a row; server should not crash with the FSEvents error.
- [ ] Existing `uv run pytest tests/ -v` still passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ wheels/
 .venv/
 .env
 uv.lock
+.oauth_clients.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ wheels/
 .venv/
 .env
 uv.lock
-.oauth_clients.json
+.oauth_clients.json*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,69 @@
+# CLAUDE.md
+
+Operational context for this repo. Code-level docs are in README.md.
+
+## How this server is actually deployed
+
+- **Runtime:** Python via `uv`, entry point `vault-mcp` (Starlette + uvicorn on port `8420`).
+- **Process manager:** launchd (`~/Library/LaunchAgents/com.marcelotoledo.vault-mcp.plist`). KeepAlive=true.
+- **Tunnel:** Tailscale Funnel, NOT Cloudflare Tunnel as the README suggests. Public URL: `https://mt-mb-pro.tailcc1eaf.ts.net`.
+- **Logs:** `~/Library/Logs/vault-mcp.log` (stdout) and `~/Library/Logs/vault-mcp-error.log` (stderr).
+- **Secrets:** `.env` at repo root (`VAULT_PATH`, `VAULT_MCP_TOKEN`, `VAULT_OAUTH_CLIENT_SECRET`, `VAULT_MCP_ALLOWED_HOSTS`). plist sources it via `set -a; . .env; set +a`.
+- **OAuth clients:** persisted to `.oauth_clients.json` so dynamically-registered clients survive restarts.
+
+### Restart
+
+```
+launchctl kickstart -k gui/$(id -u)/com.marcelotoledo.vault-mcp
+```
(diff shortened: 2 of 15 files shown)


jjsmackay:
Integrated as [jjsmackay/obsidian-web-mcp@`02fd42a`](https://github.com/jjsmackay/obsidian-web-mcp/commit/02fd42a).

jimprosser:
Thank you for the thorough work here. v0.2.0 took a different architecture for the OAuth/session layer — login-gated /oauth/authorize, mandatory S256 PKCE, fail-closed auth — and I addressed the underlying issues (#8/#29, #19, #28, #5) directly rather than merging this wholesale, since it would conflict with that design. It was a genuinely useful reference. Closing.

PR #9, opened 2026-04-12
fix: serialize datetime.date in frontmatter as ISO strings (#5)

### LINKED ISSUE #5: vault_read / vault_search / vault_batch_read fail with "Object of type date is not JSON serializable"

mleitnercom:
## Summary

Fixes #5 - `vault_read`, `vault_search`, `vault_batch_read`, and related JSON responses can crash with `Object of type date is not JSON serializable` when YAML frontmatter contains bare ISO dates like `created: 2026-04-05`.

## Root Cause

PyYAML converts bare ISO dates to `datetime.date`. Python `json.dumps()` cannot serialize these types by default.

## Fix

- Add `_DateAwareEncoder` and `vault_json_dumps` helper in `src/obsidian_vault_mcp/vault.py`.
- Replace direct `json.dumps(...)` usage in affected tool handlers with `vault_json_dumps(...)`.
- Add tests for date/datetime serialization in `tests/test_tools.py`.
- Include the Windows/ripgrep hardening and path normalization changes from the original second commit.

## Notes

This PR intentionally contains only the two original commits from the closed PR path, rebased onto upstream `main`.

### DIFF
--- a/src/obsidian_vault_mcp/frontmatter_index.py
+++ b/src/obsidian_vault_mcp/frontmatter_index.py
@@ -32,7 +32,7 @@ class FrontmatterIndex:
         for md_path in config.VAULT_PATH.rglob("*.md"):
             if self._is_excluded(md_path):
                 continue
-            rel = str(md_path.relative_to(config.VAULT_PATH))
+            rel = md_path.relative_to(config.VAULT_PATH).as_posix()
             fm = self._parse_frontmatter(md_path)
             if fm is not None:
                 self._index[rel] = fm
@@ -132,7 +132,7 @@ class FrontmatterIndex:
 
         for abs_path_str in paths:
             abs_path = Path(abs_path_str)
-            rel = str(abs_path.relative_to(config.VAULT_PATH))
+            rel = abs_path.relative_to(config.VAULT_PATH).as_posix()
             if abs_path.exists():
                 fm = self._parse_frontmatter(abs_path)
                 with self._lock:
--- a/src/obsidian_vault_mcp/tools/manage.py
+++ b/src/obsidian_vault_mcp/tools/manage.py
@@ -1,9 +1,8 @@
 """Management tools for the Obsidian vault MCP server."""
 
-import json
 import logging
 
-from ..vault import list_directory, move_path, delete_path, resolve_vault_path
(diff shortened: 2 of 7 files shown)


jimprosser:

PR #10, opened 2026-04-12
Make allowed hosts configurable via environment

mleitnercom:
## Summary

Make the DNS rebinding protection host list configurable through `VAULT_ALLOWED_HOSTS` instead of requiring source edits in `server.py`.

Today, anyone deploying behind a tunnel or reverse proxy has to patch the hard-coded `allowed_hosts` list in source and reinstall the package whenever the hostname changes. Moving this to configuration makes deployments easier to maintain without changing the default behavior.

## Changes

- add `VAULT_ALLOWED_HOSTS` parsing in `config.py`
- use `config.ALLOWED_HOSTS` in the FastMCP transport security settings
- update README examples to use the environment variable instead of editing source
- add a small config parsing test

## Default behavior

If `VAULT_ALLOWED_HOSTS` is unset, the default remains:

- `127.0.0.1:*`
- `localhost:*`
- `[::1]:*`

## Test plan

- `python -m pytest tests/test_config.py -q`
- manual smoke check with and without `VAULT_ALLOWED_HOSTS` set

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,15 +134,10 @@ export VAULT_MCP_HOSTNAME="vault-mcp.yourdomain.com"
 
 The script authenticates with Cloudflare, creates a tunnel, writes the config, and sets up the DNS record. You will need a domain managed by Cloudflare.
 
-After setup, add your tunnel hostname to the `allowed_hosts` list in `server.py` so the MCP library's DNS rebinding protection accepts requests from your domain:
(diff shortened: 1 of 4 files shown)


jimprosser:
Thank you — this is landing. Configurable allowed_hosts shipped in #34 via VAULT_MCP_ALLOWED_HOSTS. I took your minimal scope and test, but had it append to the loopback defaults rather than replace them, so localhost access can't be dropped by accident. Closing in favor of #34 — credit to you for pushing on it.

PR #11, opened 2026-04-12
Add /authorize alias for OAuth connector compatibility

mleitnercom:
## Summary

Add `/authorize` as an alias for the existing OAuth authorize handler and exempt it from bearer auth.

Some MCP connectors call `/authorize` directly instead of `/oauth/authorize`. Without this alias, the connector receives a 401 during the browser authorization step even though the discovery document points at the same logical endpoint.

## Changes

- add `/authorize` route alias to the existing OAuth authorize handler
- add `/authorize` to the bearer-auth exemption list

## Why this is separate

This PR is intentionally limited to the alias/auth-exemption compatibility fix. It does not change session handling, transport path behavior, or OAuth persistence.

## Test plan

- verify `/oauth/authorize` still works as before
- verify `/authorize` reaches the same handler instead of returning `401`

### DIFF
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -10,6 +10,7 @@ from .config import VAULT_MCP_TOKEN
 _AUTH_EXEMPT_PATHS = {
     "/health",
     "/.well-known/oauth-authorization-server",
+    "/authorize",
     "/oauth/authorize",
     "/oauth/token",
     "/oauth/register",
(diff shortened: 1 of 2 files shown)


jimprosser:
Thanks. The server advertises its endpoints via /.well-known/oauth-authorization-server, and compliant clients use the advertised /oauth/authorize + /oauth/token. A bare /authorize alias without a matching /token alias wouldn't complete the flow for a non-compliant client, so I'd rather keep the metadata as the single source of truth. Closing — appreciate the thought.

PR #13, opened 2026-04-16
Advertise OAuth discovery metadata for MCP clients

casmokey:
## Summary
- Add the RFC 9728 `/.well-known/oauth-protected-resource` endpoint so MCP clients can discover which authorization server issues tokens for this resource.
- Include a `WWW-Authenticate: Bearer resource_metadata="…"` header on 401 responses from protected endpoints, giving clients an entry point to the OAuth flow.

## Motivation
Without these, spec-compliant MCP clients (including Claude Code) have no discovery path. Instead of walking the user through OAuth, they report "Failed to connect" and never surface the authentication prompt. The existing `/.well-known/oauth-authorization-server` endpoint is necessary but not sufficient — MCP's auth spec (aligning with RFC 9728) expects the protected-resource metadata document first.

Confirmed locally: before the patch, Claude Code showed "Failed to connect" for the deployed vault server; after the patch + service restart, it walks through the OAuth flow and connects cleanly.

## Test plan
- [ ] `curl https://<host>/.well-known/oauth-protected-resource` returns JSON with `resource`, `authorization_servers`, `bearer_methods_supported`.
- [ ] `curl -I -X POST https://<host>/mcp` returns 401 with a `WWW-Authenticate: Bearer ... resource_metadata=...` header.
- [ ] Existing auth flow (bearer token on `/mcp`) still works.
- [ ] Claude Code "Add MCP server" against the deployed host walks OAuth to completion.

### DIFF
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -10,10 +10,21 @@ from .config import VAULT_MCP_TOKEN
 _AUTH_EXEMPT_PATHS = {
     "/health",
     "/.well-known/oauth-authorization-server",
+    "/.well-known/oauth-protected-resource",
     "/oauth/authorize",
     "/oauth/token",
     "/oauth/register",
 }
+
+
+def _challenge_header(request: Request, error: str) -> str:
+    base_url = str(request.base_url).rstrip("/")
+    resource_metadata = f"{base_url}/.well-known/oauth-protected-resource"
+    return (
+        f'Bearer realm="mcp", '
+        f'resource_metadata="{resource_metadata}", '
+        f'error="{error}"'
+    )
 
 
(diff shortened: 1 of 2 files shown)


jjsmackay:
Integrated as [jjsmackay/obsidian-web-mcp@`29ec18d`](https://github.com/jjsmackay/obsidian-web-mcp/commit/29ec18d).

jimprosser:
Thank you — you were right that the discovery piece was missing. The protected-resource route itself already shipped in v0.2.0 (#20), but the WWW-Authenticate challenge header on 401s was not there, and that's the part that lets a compliant client know where to authenticate. I cherry-picked exactly that into #35, crediting you. Closing in favor of #35 — appreciate you catching it.

Issue #2, opened 2026-03-21
Security: Bearer token comparison vulnerable to timing attack


ben-bl:
## Security Issue

The bearer token validation in `auth.py` (line 41) uses Python's `!=` operator:

```python
if token != VAULT_MCP_TOKEN:
```

Python's `!=` short-circuits on the first differing byte, making it vulnerable to timing side-channel attacks. Since the server is designed to be internet-facing (via Cloudflare Tunnel), an attacker who can measure response latency could theoretically recover the bearer token byte-by-byte.

## Suggested Fix

Use `hmac.compare_digest()` for constant-time comparison:

```python
import hmac

# ...

        token = auth_header[7:]
        if not hmac.compare_digest(token, VAULT_MCP_TOKEN):
            return JSONResponse(
                {"error": "Invalid token"},
                status_code=401,
            )
```

This is a one-line change plus one import. `hmac.compare_digest` is designed specifically for this purpose and is constant-time regardless of where the strings differ.

## Severity

Medium in practice (Cloudflare Tunnel adds network jitter that makes timing attacks harder), but this is a well-known security anti-pattern that should be fixed regardless.

mleitnercom:
Fix in fork: swapped token comparison to hmac.compare_digest (constant‑time). Commit: https://github.com/mleitnercom/obsidian-web-mcp/commit/bba288308b0502d063ff4535444a5996a9ff327d. Happy to open a PR if you want.

jjsmackay:
Carried in [jjsmackay/obsidian-web-mcp@`02fd42a`](https://github.com/jjsmackay/obsidian-web-mcp/commit/02fd42a) (squash of #7): `auth.py` uses `hmac.compare_digest` for the bearer token comparison.

PR #14, opened 2026-04-19
Security hardening: OAuth client auth, RFC 7591/8707 compliance, auth-bypass fixes

dr-growth:
## Summary

Hi Jim — I've been deploying your MCP server as the backbone of a personal Claude context-sharing setup (Claude Code / Desktop / web / mobile all reading and writing the same Obsidian vault through a Cloudflare Tunnel). Thanks for the clean foundation — the FastMCP + atomic-write + path-traversal work is solid.

While hardening it for an internet-exposed deployment I found a few issues worth fixing upstream. Five commits, each independent, each addressing one concern:

### 1. `Move tunnel hostname to VAULT_MCP_HOSTNAME env var`
Current code hardcodes the tunnel hostname in `allowed_hosts`. Moving it to an env var means the plist (or Docker env) can configure deployment without a code edit. Zero behavior change when env var is unset — falls back to local-only.

### 2. `Update README to reference VAULT_MCP_HOSTNAME env var`
README doc update for the change above.

### 3. `Security: constant-time token compare + remove auth-bypass fallback`
Two issues in one commit because they're in the same area:

**`auth.py`: bearer token compared with `!=`, not constant-time.**
```python
if token != VAULT_MCP_TOKEN:  # timing-attack-vulnerable
```
`oauth.py` uses `hmac.compare_digest` elsewhere for exactly this reason. The fix is a one-line swap to match.

**`server.py`: silent auth-bypass fallback.**
```python
except Exception as e:
    logger.warning(f"Could not build app ({e}), falling back to mcp.run()")
    logger.warning("Auth will NOT be enforced in this mode")
    mcp.run(transport="streamable-http", port=VAULT_MCP_PORT)
```
Any exception during auth/OAuth setup (import error, bad config, missing dep) silently drops auth and serves unauthenticated. On a tunnel exposed to the public internet that's catastrophic. Fix: remove the fallback, fail loudly.

### 4. `Security: enforce OAuth client authentication end-to-end`
Before this change, any caller reaching the tunnel could complete the OAuth dance **unauthenticated** and receive the bearer token. Specifically:

- `/oauth/authorize` reads `client_id` but never validates it against `VAULT_OAUTH_CLIENT_ID`.
- `/oauth/token` `authorization_code` grant accepts `client_id` + `client_secret` in form data but never compares them to config values (the function signature takes them but the body ignores them).

So the only real defense was hostname obscurity (DNS is public, so: none).

Changes:
- Validate `client_id` at `/authorize` with `hmac.compare_digest`.
- Validate `client_id` + `client_secret` at `/token` with `hmac.compare_digest` (mirrors how `client_credentials` already does it).
- Require PKCE `code_challenge` at `/authorize` (was optional, now mandatory, S256-only).

### 5. `Spec compliance: RFC 7591 dynamic registration + RFC 8707 resource indicators`

**RFC 7591:** I initially disabled `/oauth/register` because returning `VAULT_OAUTH_CLIENT_SECRET` to unauthenticated callers felt wrong. Turns out the MCP spec (2025-03-26) requires dynamic client registration — Claude Code's SDK actually rejects servers without it with `"Incompatible auth server: does not support dynamic client registration"`. Re-enabled the endpoint but scoped as a single-user pattern per RFC 7591 Section 3.2.1: all callers get the same pre-configured credentials. Defense-in-depth is provided by the other layers (PKCE, client-auth at `/authorize` + `/token`, bearer on every call).

**RFC 8707:** added resource indicators so tokens issued here are bound to this server's MCP URL. Prevents a token from being replayed against a different MCP server that also validates resource claims. Optional parameter — clients that don't send it get a resource-unbound token (current behavior preserved).

---

## Test plan

- [x] All existing tests pass (26/26) — no behavioral regression for legitimate clients
- [x] Claude Code `claude mcp add` with `--client-id` + `--client-secret` flags completes OAuth dance and connects to the server
- [x] Verified constant-time compares actually constant-time (swapped `==` → `hmac.compare_digest`)
- [x] Verified `/oauth/register` returns pre-configured credentials at HTTP 201 with `client_secret_expires_at: 0`
- [x] Verified auth-bypass fallback removed — server refuses to start if auth middleware can't be wired

Happy to split these into separate PRs if that's easier to review. Also happy to add unit tests for the OAuth flow paths if you want — the existing test coverage is for the vault filesystem ops, not the OAuth layer.

Background on my deployment: [dr-growth/obsidian-web-mcp](https://github.com/dr-growth/obsidian-web-mcp) (fork) + my deployment runbook [here](https://github.com/dr-growth/os/blob/main/knowledge/mcp-deployment-notes.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,16 +134,7 @@ export VAULT_MCP_HOSTNAME="vault-mcp.yourdomain.com"
 
 The script authenticates with Cloudflare, creates a tunnel, writes the config, and sets up the DNS record. You will need a domain managed by Cloudflare.
 
-After setup, add your tunnel hostname to the `allowed_hosts` list in `server.py` so the MCP library's DNS rebinding protection accepts requests from your domain:
-
-```python
-allowed_hosts=[
-    "127.0.0.1:*",
-    "localhost:*",
-    "[::1]:*",
-    "vault-mcp.yourdomain.com",  # add your hostname here
-],
-```
+Set `VAULT_MCP_HOSTNAME` to your tunnel hostname (e.g. `vault-mcp.yourdomain.com`) so the MCP library's DNS rebinding protection accepts requests from your domain. The server appends it to `allowed_hosts` automatically at startup. Leave unset for local-only use.
 
 ## Production Deployment (macOS)
 
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -1,5 +1,7 @@
 """Bearer token authentication middleware for the vault MCP server."""
 
+import hmac
+
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
 from starlette.responses import JSONResponse
@@ -37,7 +39,7 @@ class BearerAuthMiddleware(BaseHTTPMiddleware):
             )
 
         token = auth_header[7:]
-        if token != VAULT_MCP_TOKEN:
+        if not hmac.compare_digest(token, VAULT_MCP_TOKEN):
             return JSONResponse(
                 {"error": "Invalid token"},
                 status_code=401,
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -10,6 +10,10 @@ VAULT_MCP_PORT = int(os.environ.get("VAULT_MCP_PORT", "8420"))
 VAULT_OAUTH_CLIENT_ID = os.environ.get("VAULT_OAUTH_CLIENT_ID", "vault-mcp-client")
 VAULT_OAUTH_CLIENT_SECRET = os.environ.get("VAULT_OAUTH_CLIENT_SECRET", "")
 
+# Cloudflare Tunnel public hostname (e.g. vault-mcp.yourdomain.com)
+# Set this to allow the MCP library DNS rebinding protection to accept tunnel requests.
+# Leave empty for local-only use.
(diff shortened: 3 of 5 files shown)


jjsmackay:
Heads up: this looks like it conflicts with PR #7's `oauth_register` changes, which are still open upstream. PR #7 generates a unique `client_id` + per-client `client_secret` on each `/register` and persists them to `.oauth_clients.json`, and `_handle_client_credentials` already validates against that store with env-config fallback. This PR reverts `oauth_register` to return the shared `VAULT_OAUTH_CLIENT_SECRET`, which removes the per-client-secret property on the `client_credentials` path without noting it.

The mandatory PKCE + client-auth on `/authorize` and `/token` authorization_code grant are real gaps worth closing — but they could be layered on top of PR #7's model by looking up `client_id` in `_registered_clients` first (same pattern as `_handle_client_credentials`), falling back to env config. That preserves per-client-secret defense-in-depth on both grant paths instead of collapsing to a single shared secret.

For reference, we integrated PR #7 as [jjsmackay/obsidian-web-mcp@02fd42a](https://github.com/jjsmackay/obsidian-web-mcp/commit/02fd42a); happy to share the reconciled diff if helpful.

jimprosser:
Thank you for the detailed security pass. The core hardening — constant-time token compare, redirect_uri allowlisting, client_id binding, mandatory PKCE, fail-closed auth — all shipped in v0.2.0 (#2/#4/#8/#29), implemented to keep dynamic client registration non-leaking. Merging this as-is would reintroduce the configured client secret in the registration response, so I'm closing it — but we landed at the same place on the fundamentals, and I appreciate the careful review.

PR #17, opened 2026-04-24
feat(write): block Markdown checkboxes by default to protect Task Forge users

takeru1984:
## Summary

Adds a `allow_checkboxes: bool = False` parameter to `vault_write`. By default, writes that contain Markdown checkboxes (`- [ ]`, `* [x]`, `+ [ ]`, …) outside fenced code blocks are rejected with a JSON error explaining how to opt in.

## Why

In vaults that use the [Task Forge](https://github.com/zachatoo/obsidian-task-forge) plugin (and other plugins that aggregate `- [ ]` markers as tasks), every checkbox an LLM writes becomes a tracked task. Models love sprinkling "next steps" / "follow-ups" checkboxes into ordinary meeting notes, knowledge dumps, and drafts — and once the task list balloons with phantom items, you have to clean them up by hand.

Prompt-level rules ("don't use checkboxes in plain notes") get broken regularly. Enforcing it server-side is the only durable fix.

## Behavior

| `allow_checkboxes` | Content has checkbox | Result |
|---|---|---|
| `False` (default) | No | normal write |
| `False` (default) | Yes | rejected, JSON error returned |
| `True` | No | normal write |
| `True` | Yes | normal write (intentional task list) |

Fenced code blocks (` ```…``` `) are stripped before the check so docs / examples that demonstrate checkbox syntax are not blocked. Inline code (`` ` ``) is left as a TODO — easy to extend later if false positives appear.

The tool description tells the LLM the rule explicitly: `allow_checkboxes=True` requires **explicit** user task-creation intent ("タスク登録して", "TODO リストにして", etc.), never just because checkboxes "would look nice".

## Backward compatibility

Pure additive: existing callers passing only `path` / `content` / `create_dirs` / `merge_frontmatter` continue to work unchanged unless their content already contained checkboxes — which under Task Forge was already producing phantom tasks.

## Test plan

9 new tests in `tests/test_tools.py`:
- blocks-by-default
- allows-when-opted-in
- detects all bullet × state variants (`- [ ]` / `- [x]` / `- [X]` / `* [ ]` / `* [x]` / `+ [ ]` / `+ [X]` / indented / tab-indented)
- ignores checkboxes inside fenced code blocks
- plain markdown with bullets/numbered lists still succeeds

`pytest tests/test_tools.py` → 20 passed locally (11 existing + 9 new).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,32 @@ This is a server that provides network access to your personal notes. Security i
 |------|-------------|
 | `vault_read` | Read a file, returning content, metadata, and parsed YAML frontmatter |
 | `vault_batch_read` | Read multiple files in one call; handles missing files gracefully |
-| `vault_write` | Write a file with optional frontmatter merging; creates parent dirs |
+| `vault_write` | Write a file with optional frontmatter merging; creates parent dirs. Blocks Markdown checkboxes (`- [ ]`) by default — set `allow_checkboxes=True` only when the user explicitly asked for tasks (see "Checkbox guard" below) |
 | `vault_batch_frontmatter_update` | Update YAML frontmatter fields on multiple files without touching body content |
 | `vault_search` | Full-text search across vault files (uses ripgrep if available, falls back to Python) |
 | `vault_search_frontmatter` | Query the in-memory frontmatter index by field value, substring, or field existence |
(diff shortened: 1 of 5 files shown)


takeru1984:
Closing — opened in error without owner approval. Keeping the change in our internal fork only.

PR #21, opened 2026-05-20
fix: serialize datetime.date frontmatter fields as ISO 8601 strings

MerchGooSe:
## Problem

When a vault note has an unquoted YAML date field in its frontmatter — e.g.

```yaml
---
created: 2026-05-18
updated: 2026-05-18
---
```

— PyYAML (via `python-frontmatter`) automatically parses these as `datetime.date` objects. `json.dumps()` then fails with:

```
TypeError: Object of type date is not JSON serializable
```

This breaks every tool that returns frontmatter to the client: `vault_read`, `vault_batch_read`, `vault_search`, `vault_search_frontmatter` — effectively making the server unusable for any vault containing notes with unquoted date frontmatter.

This is a fairly common pattern in Obsidian vaults since the popular [Templater](https://github.com/SilentVoid13/Templater) and Daily Notes plugins both emit unquoted dates by default.

## Solution

New module `src/obsidian_vault_mcp/json_utils.py` exposes a `dumps()` helper that wraps `json.dumps` with a `default=` handler converting `date` / `datetime` / `time` to ISO 8601 strings:

```python
def _default(obj):
    if isinstance(obj, (date, datetime, time)):
        return obj.isoformat()
    raise TypeError(
        f"Object of type {type(obj).__name__} is not JSON serializable"
    )

def dumps(data) -> str:
    return json.dumps(data, default=_default)
```

All four tool modules (`read.py`, `write.py`, `search.py`, `manage.py`) now import this helper instead of calling `json.dumps` directly.

## Why this approach

- **No content-side workaround required** — users can keep writing unquoted dates in their frontmatter
- **Minimal blast radius** — only the serialization boundary changes, YAML parsing logic is untouched
- **Same Python stdlib pattern** as `json.JSONEncoder` subclassing, just simpler with `default=`
- **Forward-compatible** — easy to extend for other non-serializable types if needed

## Tests

New test file `tests/test_json_utils.py` with 8 unit tests covering:

- Plain string serialization (regression)
- `datetime.date` → ISO string
- `datetime.datetime` → ISO timestamp
- `datetime.time` → ISO time
- Nested dates in lists/dicts
- Realistic frontmatter payload (mimics `python-frontmatter` output)
- `TypeError` still raised for unsupported types
- `None` / `True` / `False` passthrough

All existing tests (26) continue to pass. Full suite is **34/34 green**.

## Verification

Tested live against a vault with 158+ notes, several with unquoted `created:` / `updated:` / `start:` / `ende:` date fields. Before the fix: `vault_read` returned `{"error": "Object of type date is not JSON serializable"}`. After: returns proper JSON with date fields as `"2026-05-18"` ISO strings.

## Commits

1. `9bac2f0` — `fix(serialization): handle datetime.date in JSON responses`
2. `5c7684f` — `test(json_utils): add unit tests for date/datetime/time serialization`

### DIFF
--- a/src/obsidian_vault_mcp/json_utils.py
+++ b/src/obsidian_vault_mcp/json_utils.py
@@ -1,0 +1,23 @@
+"""JSON serialization helpers for vault MCP tools.
+
+PyYAML / python-frontmatter auto-convert unquoted YAML dates
+(e.g. `created: 2026-05-18`) to datetime.date objects, which json.dumps
+cannot serialize by default. This helper converts date/datetime/time
+objects to ISO 8601 strings so vault_read and friends return proper JSON.
+"""
+import json
+from datetime import date, datetime, time
+
+
+def _default(obj):
+    """Convert non-JSON-serializable types to JSON-safe representations."""
+    if isinstance(obj, (date, datetime, time)):
+        return obj.isoformat()
+    raise TypeError(
+        f"Object of type {type(obj).__name__} is not JSON serializable"
+    )
+
+
+def dumps(data) -> str:
+    """json.dumps that handles datetime.date / datetime / time objects."""
+    return json.dumps(data, default=_default)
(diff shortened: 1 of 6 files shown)


jimprosser:

PR #22, opened 2026-05-25
fix: handle YAML date objects in vault_search and vault_read

Sinc3today:
## Summary

`vault_search`, `vault_search_frontmatter`, `vault_read`, and `vault_batch_read` raise `TypeError: Object of type date is not JSON serializable` (surfaced to MCP clients as an `isError` tool result) when frontmatter contains unquoted ISO 8601 dates such as `created: 2026-05-25`.

PyYAML — used under the hood by `python-frontmatter` — parses unquoted ISO dates as `datetime.date` objects. These objects flow into the `json.dumps()` response builders in `tools/search.py` and `tools/read.py`, which have no `default=` fallback, so the date object bubbles up as a `TypeError`.

## Fix

Add a small `_json_default` helper to `tools/search.py` and `tools/read.py` that converts `datetime.date`, `datetime.datetime`, and `datetime.time` to ISO 8601 strings at serialization time. Anything else still raises, so other type bugs are not silently swallowed.

## Reproduction (before patch)

1. Create a note with unquoted ISO date frontmatter:

   ```markdown
   ---
   created: 2026-05-25
   tags: [test]
   ---
   body
   ```

2. Call `vault_search` for any term matching the body, or `vault_read` on that file.
3. The tool returns an error containing `Object of type date is not JSON serializable`.

## Test plan

- [x] Created vault file with unquoted `created: 2026-05-25` in frontmatter
- [x] Confirmed `vault_search` returns clean JSON with `frontmatter_excerpt: {"created": "2026-05-25", ...}`
- [x] Confirmed `vault_read` returns the file with frontmatter serialized as ISO string
- [x] Confirmed `vault_search_frontmatter` returns clean JSON (also affected)
- [x] `python -m py_compile` clean on both modified modules
- [ ] Existing pytest suite not run on this branch — `pytest` is in optional `[dev]` extras and was not installed in the deployment venv used to verify this patch. Patch surface is small (adds `default=` callable only, no logic changes), so regression risk is low. Happy to run the suite if useful.

### DIFF
--- a/src/obsidian_vault_mcp/tools/read.py
+++ b/src/obsidian_vault_mcp/tools/read.py
@@ -1,5 +1,6 @@
 """Read tools for the Obsidian vault MCP server."""
 
+import datetime
 import json
 import logging
 
@@ -8,6 +9,12 @@ import frontmatter
 from ..vault import resolve_vault_path, read_file
 
 logger = logging.getLogger(__name__)
+
+
+def _json_default(obj):
+    if isinstance(obj, (datetime.date, datetime.datetime, datetime.time)):
+        return obj.isoformat()
+    raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
 
 
 def vault_read(path: str) -> str:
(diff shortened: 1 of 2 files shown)


jimprosser:

PR #24, opened 2026-05-27
Add VAULT_MCP_ALLOWED_HOSTS + Start script

tleyden:
I got it running on my VPS and connected to Claude Desktop + Mobile (yay!) and made a few tweaks along the way:

- Added `VAULT_MCP_ALLOWED_HOSTS` to make it easier to extend allowed hosts without having to hardcode them.  This seems pretty useful, and if needed I can extract this change into a separate PR
- Start script that does some validation
- Add `.env.template` to make it easier to copy and edit
- Update `README` with instructions on running `server-start`, as well as manually verifying the OAuth2 via `curl`.  Not sure how useful this is though, something like [MCPJam Inspector](https://github.com/MCPJam/inspector) is probably a better approach.

Let me know if you want any of these cleaned up and split out into separate PRs!

# AI Generated Summary

  - Start script (scripts/start-server.sh) — single command to launch the server; auto-sources .env; validates required secrets are set; prompts to run uv sync if virtualenv is missing; warns when
  VAULT_MCP_ALLOWED_HOSTS is unset so it's clear the server is localhost-only
  - Secret generation (scripts/generate_secrets.sh) — generates cryptographically secure tokens for VAULT_MCP_TOKEN and VAULT_OAUTH_CLIENT_SECRET, writes them to .env (mode 600, gitignored), guards against
  accidental overwrite
  - Remote client support — new VAULT_MCP_ALLOWED_HOSTS env var (comma-separated) extends the DNS rebinding protection allowlist, fixing 421 errors when connecting from remote MCP clients like Claude web or MCPJam
  - Config template (.env.template) — documents all env vars with inline comments and example values; secrets blanked out for safe copying
  - README — adds setup and server-start instructions; adds curl-based smoke test for verifying auth and tool listing; updates Configuration table with all vars including VAULT_MCP_ALLOWED_HOSTS

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,0 +1,19 @@
+# Required: path to your Obsidian vault directory
+VAULT_PATH=/path/to/your/obsidian_vaults
+
+# Required: generate with: python3 -c "import secrets; print(secrets.token_hex(32))"
+VAULT_MCP_TOKEN=
+
+# Required: generate with: python3 -c "import secrets; print(secrets.token_hex(32))"
+VAULT_OAUTH_CLIENT_SECRET=
+
+# Optional: OAuth client ID (default: vault-mcp-client)
+VAULT_OAUTH_CLIENT_ID=vault-mcp-client
+
+# Optional: port to listen on (default: 8420)
+VAULT_MCP_PORT=3001
+
+# Optional: comma-separated extra allowed Host headers for DNS rebinding protection
+# Required if connecting from a remote MCP client (e.g. MCPJam, Claude web)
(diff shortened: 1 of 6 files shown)


jimprosser:
Thanks — configurable allowed_hosts landed in #34 (VAULT_MCP_ALLOWED_HOSTS, appended to the loopback defaults), credit to you and #10 for pushing on it. I kept it to the env var for now and left the start-server/secrets tooling separate. Closing in favor of #34; appreciate the contribution.

PR #25, opened 2026-05-27
VPS setup instructions

tleyden:
Added README instructions for VPS setup with Cloudflare TLS + Caddy.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,118 @@ scripts/
     launchd/                # macOS launchd plist templates
 ```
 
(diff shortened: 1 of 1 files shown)


jimprosser:
Thanks — verified the steps against current main: the uv run vault-mcp entry point, port 8420, serving MCP at /, and the allowed_hosts edit are all correct. Merging. Heads-up: configurable allowed_hosts is landing shortly, after which step 5 becomes an env var rather than a server.py edit — I'll update this section then. Appreciate the clear write-up.

PR #27, opened 2026-06-02
Add edit tools and harden vault response serialization

ProfessionalSeaweedDevourer:
## Summary

Adds two token-efficient write-side tools, rebased onto current `main`:

- `vault_edit`: patch an existing file with ordered exact text replacements, with optional dry-run unified diff previews.
- `vault_append`: append content to the end of a file, or create the file when it does not exist, without resending the existing body.

Per review feedback, this revision is narrowed to just these two tools and their tests. The earlier serialization changes are dropped, since v0.2.0 already routes responses through the shared `serialization.py` helper. The README auth section is left untouched; only the `## Tools` table gains the two new rows.

## Motivation

This targets the documented remote connector shape: the MCP server runs locally, a Cloudflare Tunnel forwards the HTTPS endpoint, and Claude.ai connects as a remote MCP connector. In that workflow, small note changes currently require sending the full Markdown file through `vault_write`, which spends far more context and tokens than the actual edit needs, especially for daily notes, instructions, and logs.

## Behavior

`vault_edit`:

- Requires every `old_text` fragment to appear exactly once in the current file content.
- Returns an error without writing when a fragment is missing or appears more than once.
- Applies multiple edits sequentially in memory, then writes once only if every edit matches (a later failing edit leaves the file untouched).
- Supports `dry_run=true`, returning a unified diff preview without changing the file.
- Accepts `old_text`/`new_text`, plus compatible `old_str`/`new_str` aliases for str_replace-style clients, and rejects mixing a canonical key with its alias in the same edit.

`vault_append`:

- Appends content to an existing file using a configurable separator.
- Creates a missing file when `create_dirs=true`, matching the default write behavior.
- Is intentionally simple and append-only; precise section insertion should use `vault_edit`.

Both tools register with write-side MCP annotations: `readOnlyHint=false`, `destructiveHint=true`, `idempotentHint=false`, `openWorldHint=false`.

## Tests

Adds regression tests for `vault_edit` (single match, dry-run diff, str_replace aliases, mixed-key rejection, missing/repeated fragment, atomic rollback) and `vault_append` (append, create, code-fence preservation), plus patch-style input model tests. Full suite green (69 passed).

## Safety and Compatibility

Existing `vault_write`, move, delete, search, and frontmatter update semantics are preserved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ Found a vulnerability? Please report it privately rather than opening a public i
 | `vault_read` | Read a file, returning content, metadata, and parsed YAML frontmatter |
 | `vault_batch_read` | Read multiple files in one call; handles missing files gracefully |
 | `vault_write` | Write a file with optional frontmatter merging; creates parent dirs |
+| `vault_edit` | Patch a file with ordered exact text replacements (token-efficient partial edits); supports dry-run diff previews |
+| `vault_append` | Append content to the end of a file without resending the existing body; creates the file when missing |
 | `vault_batch_frontmatter_update` | Update YAML frontmatter fields on multiple files without touching body content |
 | `vault_search` | Full-text search across vault files (uses ripgrep if available, falls back to Python) |
 | `vault_search_frontmatter` | Query the in-memory frontmatter index by field value, substring, or field existence |
--- a/src/obsidian_vault_mcp/models.py
+++ b/src/obsidian_vault_mcp/models.py
@@ -2,7 +2,7 @@
 
 from typing import Literal
 
-from pydantic import BaseModel, ConfigDict, Field, field_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from .config import (
     CONTEXT_LINES,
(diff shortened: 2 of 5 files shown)


jimprosser:
Thank you for this — vault_edit and vault_append are genuinely useful, and token-efficient edits from remote clients are squarely where I want this server to go. Two things before I can merge: (1) v0.2.0 added a shared serialization.py encoder that all tool responses now route through, so the serialization changes here conflict — could you rebase onto current main and drop them, keeping just the vault_edit/vault_append tools and their tests? (2) the README auth section is now out of date (the server login-gates /oauth/authorize and issues per-client credentials). Rebase down to just the two tools and I'll review and merge. Really appreciate the contribution.

ProfessionalSeaweedDevourer:
Thanks for the review! Rebased onto current `main` and narrowed the PR down to just `vault_edit` and `vault_append` plus their tests:

- Dropped the standalone serialization changes (and `serialization.py`), since v0.2.0 already routes responses through the shared encoder.
- Left the README auth/security section untouched; only added the two new rows to the `## Tools` table.

It is now a single clean commit on top of current `main` (5 files, +490/-3). Full test suite is green (69 passed). PR description updated to match the narrowed scope.

PR #38, opened 2026-06-08
perf: emit non-ASCII tool responses as UTF-8 instead of \uXXXX escapes

ProfessionalSeaweedDevourer:
> **Foundation of a stack.** #40 (vault_edit diagnostics) builds on this PR's `serialization.dumps` change in `write.py`, so this one merges first. #39 (ripgrep context_lines) is independent of both.

## What

Tool responses go through `serialization.dumps`, but it inherited json.dumps's
default `ensure_ascii=True`, which escapes every non-ASCII character into a
`\uXXXX` sequence (a single non-BMP emoji becomes a 12-character surrogate pair).
Two tool modules (`manage.py`, `write.py`) also called `json.dumps` directly,
bypassing the shared encoder.

This PR:

- Sets `ensure_ascii=False` and compact `separators` on `serialization.dumps`
  (callers may still override either via kwargs).
- Routes `manage.py` and `write.py` through `serialization.dumps` so one encoder
  governs every response. `read.py` and `search.py` already used the shared
  encoder, so read and search responses gain the same verbatim, compact output
  too. `frontmatter.dumps` calls are left untouched.
- Guards the encoder against non-UTF-8 (surrogate-escaped) filenames: `dumps`
  verifies its output is UTF-8 encodable and falls back to escaped output
  otherwise, so a single invalid filename cannot crash a response at the
  transport layer.
- Removes a pre-existing unused import in `manage.py` flagged by ruff.

## Why

For any vault whose content is not pure ASCII, escaping inflates every response.
Byte size on representative single-note tool responses (old `ensure_ascii=True`
-> new):

| Script | search response | read response |
|--------|----------------:|--------------:|
| Korean | 426 -> 282 (34%) | 717 -> 434 (39%) |
| Japanese | 573 -> 351 (39%) | 822 -> 476 (42%) |
| Chinese | 429 -> 279 (35%) | 582 -> 356 (39%) |
| Cyrillic | 665 -> 318 (52%) | 1262 -> 525 (58%) |
| Arabic | 587 -> 292 (50%) | 1052 -> 455 (57%) |
| Devanagari | 473 -> 308 (35%) | 872 -> 517 (41%) |
| Greek | 606 -> 299 (51%) | 1197 -> 504 (58%) |
| Hebrew | 531 -> 272 (49%) | 927 -> 410 (56%) |
| Thai | 735 -> 432 (41%) | 1092 -> 611 (44%) |
| Accented Latin | 274 -> 235 (14%) | 392 -> 335 (15%) |
| Emoji (non-BMP) | 271 -> 224 (17%) | 417 -> 324 (22%) |
| English (pure ASCII) | 248 -> 233 (6%) | 347 -> 334 (4%) |

The token-count reduction is larger still, since `\uXXXX` escapes tokenize into
several tokens each. This compounds across every read, search, list, and write
response, and grows with document size (see the scale numbers under Tests).

Beyond size, in day-to-day use of a Korean-language vault the escaping also
caused practical breakage: a path a client echoed back (for a follow-up read or
edit) did not always match the stored file once it had been round-tripped
through the escaped form, which surfaced as duplicate notes and missed search
hits. Emitting paths and content verbatim removes that class of problem.

## Safety

Behavior-preserving: the decoded object is identical to the old escaped output
(`json.loads(old) == json.loads(new) == obj`) for every script in the table.
`ensure_ascii=False` output is valid JSON and MCP responses already travel as
UTF-8 over HTTP. Pure-ASCII responses are unchanged apart from separator
whitespace.

## Tests

Two new test modules; full suite green (**176 passed**), ruff clean on touched
files.

### Correctness across scripts (`tests/test_serialization_encoding.py`)

Parametrizes 11 scripts (Korean, Japanese, Chinese, Cyrillic, Arabic, Devanagari,
Greek, Hebrew, Thai, accented Latin, non-BMP emoji) and asserts end-to-end
through `vault_read`, `vault_write`, `vault_list`, `vault_search`, and the
`vault_edit` dry-run diff that responses are emitted verbatim (no `\uXXXX`),
decode identically to the old encoder, and are strictly smaller. Additional cases
cover `vault_search_frontmatter`, `vault_move`/`vault_delete`, `vault_append`, non-ASCII
error-path responses, the kwargs override contract, the non-UTF-8 filename guard,
a #5 date-serialization regression guard, and an ASCII-unchanged control.

### Real-world scale (`tests/test_serialization_scale.py`)

Uses authentic public-domain text (Universal Declaration of Human Rights
excerpts across ten scripts, see `tests/fixtures/udhr/README.md`). Each test
runs the real tool and asserts verbatim output, exact round-trip integrity of
the multilingual body, and a meaningful size reduction. Measured on the actual
tool responses (old `ensure_ascii=True` -> new):

| Scenario (real tool response) | old | new | reduction | saved |
|-------------------------------|----:|----:|----------:|------:|
| `vault_read`, 270-line multilingual note | 156,602 B | 72,386 B | 54% | ~82 KB |
| `vault_batch_read`, 10 scripts | 40,924 B | 19,789 B | 52% | ~21 KB |
| `vault_search`, many matches in a large file | 94,022 B | 51,411 B | 45% | ~42 KB |
| `vault_edit` dry-run diff, 30 edits | 23,790 B | 12,524 B | 47% | ~11 KB |
| `vault_batch_frontmatter_update`, 10 files | 433 B | 393 B | 9% | 40 B |

The last row is small because that response carries status, not content; the
saving lands on content-bearing responses, and grows with document size. The
`vault_edit` dry-run row matters specifically for the token-efficient edit tools:
the diff preview is no longer escaped.

Running `pytest tests/test_serialization_encoding.py tests/test_serialization_scale.py`
reproduces all of the above.


ProfessionalSeaweedDevourer:
Heads-up on changes since this PR was opened (delivered as follow-up commits on this branch, not just the original commit):

**1. Test suite expanded** from the initial smoke coverage to:
- per-language correctness across 11 scripts (CJK, Cyrillic, Arabic, Devanagari, Greek, Hebrew, Thai, accented Latin, non-BMP emoji), end-to-end through `vault_read`/`vault_write`/`vault_list`/`vault_search` and the `vault_edit` dry-run diff;
- real-world-scale tests using public-domain UDHR excerpts (hundreds-of-lines documents, bulk edit, batch read, batch frontmatter update, many-match search), asserting verbatim output, round-trip integrity, and size reduction on the actual tool responses.

**2. Self-review hardening** (incoming follow-up commit):
- guard the shared encoder against non-UTF-8 (surrogate-escaped) filenames, so one odd filename on disk cannot turn a `vault_list` into a transport-level `UnicodeEncodeError`;
- added tests for `vault_search_frontmatter`, `vault_move`/`vault_delete`, batch-read missing files, non-ASCII error-path responses, and the `kwargs` override contract;
- documentation accuracy fixes.

One note for review: `read.py` and `search.py` already routed through the shared encoder, so changing the defaults there also makes read/search responses non-ASCII-verbatim and compact, not only `manage.py`/`write.py`.

Happy to squash the branch to a single commit before merge if you prefer that to the follow-up commit.

ProfessionalSeaweedDevourer:
Opened #49 to capture the user-facing motivation for this change (token/byte inflation and path round-trip breakage on non-ASCII vaults), independent of the implementation here. Refs #49.

jimprosser:
Reviewed this closely (independent second pass included). The encoding is handled at the right layer — every tool returns a `str` and the Streamable HTTP transport owns the wire encoding with a byte-correct `Content-Length` and `charset=utf-8`, so the usual `ensure_ascii=False` failure modes don't apply here; I confirmed nothing in the repo hand-frames responses or computes `Content-Length` manually (`json_response=True`). The lone-surrogate guard for non-UTF-8 filenames is the right defensive touch, and the diff is mostly tests/fixtures rather than logic churn. No security regression — JSON still escapes quotes/control chars regardless of `ensure_ascii`.

LGTM. I'll squash-merge. Thanks — and the path round-trip fix (not just the size win) is the real value here.

PR #39, opened 2026-06-08
fix: honor context_lines consistently in vault_search (ripgrep vs Python)

ProfessionalSeaweedDevourer:
## Problem

`vault_search` returns differently shaped results depending on whether `ripgrep` is installed.

`_search_ripgrep` passes `--context={context_lines}` to `rg --json`, but the parsing loop only handles `type == "match"` events and silently discards the `type == "context"` events that ripgrep emits. So on the ripgrep path `match_context` is just the single matched line, while the Python fallback (`_search_python`) returns the matched line plus `context_lines` lines on each side.

Net effect:
- `context_lines` (documented param, default `2`) is effectively a no-op on the ripgrep path.
- The same query against the same vault returns different `match_context` shapes depending on whether `rg` is on `PATH`.

The field name `match_context` and the default `context_lines=2` both indicate surrounding context was intended.

### Reproduce

```bash
printf 'line one\nTARGET here\nline three\n' > /tmp/rgtest.md
rg --json --context=2 TARGET /tmp/rgtest.md | python3 -c "import sys,json
for l in sys.stdin:
    d=json.loads(l)
    if d.get('type') in ('match','context'):
        print(d['type'], repr(d['data']['lines']['text'].rstrip()))"
# rg emits context/match/context events, but the code only consumed 'match'
```

## Fix

Make both backends share one context-assembly path:

- Extract the slice logic the Python backend already used into `_assemble_context(file_lines, match_index, context_lines)`.
- The ripgrep backend now takes only the line number from `rg`, reads the matched file (cached per file so repeated matches in one file read once), and assembles context with the same helper.
- Drop the now-unused `--context` flag from the `rg` invocation.

Both backends produce identical `match_context` for the same query, and `context_lines` works on both paths. No change to `frontmatter_excerpt`, `--max-count`, or the overall result cap.

### Line counting and encoding

To keep the re-read context aligned with ripgrep's `line_number`, both backends read raw bytes (`read_bytes().decode("utf-8")`, not text mode) and split with a shared `_split_lines()` that breaks on `\n` only, the same way ripgrep counts lines. This avoids two subtle off-by-N bugs:

- Text-mode reading translates a lone `\r` into `\n`, and `str.splitlines()` also breaks on other Unicode separators (NEL, LINE/PARAGRAPH SEPARATOR, etc.) that ripgrep does not. Either would shift the re-read index off ripgrep's line number and drop or mis-center context.
- Trailing carriage returns are stripped so CRLF files render the same as `str.splitlines()` output.

If a matched file cannot be re-read (deleted between the rg scan and the read, or invalid UTF-8), the ripgrep backend logs a warning and degrades to the matched line ripgrep already returned, base64-decoding ripgrep's `bytes` payload for non-UTF-8 lines so the snippet is never empty or misindexed.

This is a narrow, single-purpose bug fix. It does not change the response shape for callers that relied on the documented multi-line context (the Python path already behaved this way); it brings the ripgrep path into line with it.

## Tests

New `tests/test_search_context.py`:
- both backends honor `context_lines` (multi-line snippet)
- backend equivalence: ripgrep and Python paths return identical `match_context` for the same query (forced via `shutil.which` monkeypatch)
- `context_lines=0` returns only the matched line
- clamping at the first / last line of a file
- multiple matches in one file (exercises the per-file read cache)
- the read-failure fallback degrades to the matched line, not an empty string
- Unicode line separators (U+2028), lone `\r`, and CRLF keep both backends' line indexing aligned
- non-UTF-8 matched lines produce a non-empty snippet
- the default `context_lines` value (2) is wired through
- both backends agree on `total_matches` and `truncated` for a capped multi-file search

Full suite green (83 passed), lint clean. ripgrep-path tests skip automatically when `rg` is not installed.

## Known backend differences (out of scope)

These predate this PR and are intentionally left for a separate follow-up so this change stays a narrow context_lines fix:

- **Query semantics**: ripgrep treats the query as a regex while the Python fallback does a literal case-insensitive substring match, so a query with regex metacharacters can match differently between backends.
- **Non-UTF-8 files**: the Python fallback skips a file it cannot decode, while ripgrep still returns a (degraded) match for it.
- **Match ordering** across multiple files differs (ripgrep stdout order vs Python `rglob` order); `total_matches` and `truncated` agree, only the order can differ.
- **Backend-level silent failures** in `_search_ripgrep` (subprocess timeout / missing binary / nonzero `rg` exit code return an empty list with no log) and the broad `except` blocks in `vault_search` are pre-existing and unchanged here.

### DIFF
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -81,6 +81,7 @@ MAX_SEARCH_RESULTS = 50       # Max results per search
 DEFAULT_SEARCH_RESULTS = 20
 MAX_LIST_DEPTH = 5            # Max directory recursion depth
 CONTEXT_LINES = 2             # Default lines of context in search results
+MAX_SEARCH_REREAD_BYTES = MAX_CONTENT_SIZE  # Cap per-file context re-read in vault_search
 
 # Directories to never expose or modify
 EXCLUDED_DIRS = {".obsidian", ".trash", ".git", ".DS_Store"}
--- a/src/obsidian_vault_mcp/tools/search.py
+++ b/src/obsidian_vault_mcp/tools/search.py
@@ -3,6 +3,7 @@
 import base64
 import json
 import logging
+import os
 import shutil
 import subprocess
 from pathlib import Path
@@ -65,6 +66,100 @@ def _assemble_context(
     return "\n".join(file_lines[start:end])
 
 
+def _open_in_vault_nofollow(vault_root: Path, rel_path: str) -> int:
+    """Open vault_root/rel_path for reading without following a symlink in any
+    path component below the (trusted, configured) vault root.
+
+    The matched path comes back from a separate ripgrep process and is opened in
+    a later syscall, so a single path-based open is a TOCTOU: a symlink swapped
+    into an intermediate component between a containment check and the open would
+    be followed. Here each component under the root is opened with O_NOFOLLOW
+    relative to its parent's directory fd (openat), so a symlink anywhere (final
+    or intermediate, even one swapped in concurrently) fails the open atomically
+    rather than being resolved by a racy path lookup. The configured root itself
+    is opened normally, since it is allowed to be a symlink (VAULT_PATH may point
+    at, e.g., a mounted volume).
+
+    Where openat/dir_fd is unavailable (e.g. Windows) this degrades to a single
+    O_NOFOLLOW open on the joined path, guarded by a (racy) resolved-path
+    containment check. Raises OSError on any refusal so the caller degrades to
+    ripgrep's matched line.
+    """
+    nofollow = getattr(os, "O_NOFOLLOW", 0)
+    directory = getattr(os, "O_DIRECTORY", 0)
+    parts = [p for p in Path(rel_path).parts if p not in ("", os.curdir)]
+
+    if os.open not in os.supports_dir_fd or not parts:
+        full = os.path.join(str(vault_root), rel_path)
+        resolved = os.path.realpath(full)
+        root = os.path.realpath(str(vault_root))
+        if resolved != root and not resolved.startswith(root + os.sep):
+            raise OSError("matched path resolves outside the vault")
+        return os.open(full, os.O_RDONLY | nofollow)
+
+    dir_fd = os.open(str(vault_root), os.O_RDONLY | directory)
+    try:
+        for part in parts[:-1]:
+            next_fd = os.open(part, os.O_RDONLY | nofollow | directory, dir_fd=dir_fd)
+            # Reassign before closing so an interrupted close (EINTR) leaves the
+            # finally clause closing the fd we actually still hold, not a stale one.
+            old_fd, dir_fd = dir_fd, next_fd
+            os.close(old_fd)
+        return os.open(parts[-1], os.O_RDONLY | nofollow, dir_fd=dir_fd)
+    finally:
+        os.close(dir_fd)
+
+
+def _read_vault_file_lines(rel_path: str) -> list[str] | None:
+    """Re-read a ripgrep-matched vault file for context, refusing unsafe reads.
+
+    rel_path is the match's path relative to the vault root (already validated as
+    lexically inside it by the caller). Returns the file's lines (per
+    _split_lines), or None if the file cannot be read safely, in which case the
+    caller degrades to ripgrep's matched line.
+
+    Two guards, because ripgrep reports the path from a separate process and we
+    open it later:
+      * Symlinks are never followed below the vault root (see
+        _open_in_vault_nofollow), so a matched note that is, or sits under, a
+        symlink pointing out of the vault cannot leak outside content, even if
+        the symlink is swapped in concurrently.
+      * The size cap bounds memory: the file is read in full to slice context, so
+        without a limit a few huge notes (or a broad query) could spike memory
+        despite the result cap. Oversized files degrade to the matched line.
+
+    Not caught here: a hardlink inside the vault to an outside file (a hardlink is
+    a real directory entry, indistinguishable from a normal file). That is a
+    pre-existing, vault-wide exposure shared with vault.read_file, not specific to
+    search.
+    """
+    try:
+        fd = _open_in_vault_nofollow(config.VAULT_PATH, rel_path)
+        try:
+            if os.fstat(fd).st_size > config.MAX_SEARCH_REREAD_BYTES:
+                raise OSError("matched file too large to re-read for context")
+            chunks = []
+            while True:
+                block = os.read(fd, 65536)
+                if not block:
+                    break
+                chunks.append(block)
+        finally:
+            os.close(fd)
+
+        # read_bytes-equivalent, not read_text: text mode translates lone "\r"
+        # to "\n", which would desync _split_lines from rg's line count.
+        return _split_lines(b"".join(chunks).decode("utf-8"))
+    except (OSError, UnicodeDecodeError, ValueError) as exc:
+        logger.warning(
+            "vault_search: could not safely re-read %s for context, "
+            "falling back to matched line only (%s)",
+            rel_path,
+            exc,
+        )
+        return None
+
+
 def _search_ripgrep(
     query: str,
     search_path: Path,
@@ -119,25 +214,18 @@ def _search_ripgrep(
             line_number = match_data["line_number"]
 
             if file_path not in file_lines_cache:
-                try:
-                    # read_bytes, not read_text: text mode translates lone "\r"
-                    # to "\n", which would desync _split_lines from rg's count.
-                    file_lines_cache[file_path] = _split_lines(
-                        Path(file_path).read_bytes().decode("utf-8")
-                    )
-                except (OSError, UnicodeDecodeError) as exc:
-                    logger.warning(
-                        "vault_search: could not re-read %s for context, "
-                        "falling back to matched line only (%s)",
-                        file_path,
-                        exc,
-                    )
-                    file_lines_cache[file_path] = None
+                file_lines_cache[file_path] = _read_vault_file_lines(rel_path)
 
             file_lines = file_lines_cache[file_path]
-            if file_lines is not None:
+            match_index = line_number - 1
+            # Degrade to ripgrep's matched line when the re-read was refused
+            # (file_lines is None) or when rg's line_number is past the file's
+            # current length -- a content race where the file shrank between rg's
+            # scan and the re-read -- which would otherwise slice to an empty
+            # snippet instead of the matched line.
+            if file_lines is not None and 0 <= match_index < len(file_lines):
                 match_context = _assemble_context(
-                    file_lines, line_number - 1, context_lines
+                    file_lines, match_index, context_lines
                 )
             else:
                 match_context = _rg_match_line(match_data)
--- a/src/obsidian_vault_mcp/tools/search.py
+++ b/src/obsidian_vault_mcp/tools/search.py
@@ -1,5 +1,6 @@
 """Search tools for the Obsidian vault MCP server."""
 
+import base64
 import json
 import logging
 import shutil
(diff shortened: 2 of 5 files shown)


jimprosser:
The `context_lines` consistency fix is correct and well-tested — routing both backends through one snippet assembler is the right call, the 1-based→0-based ripgrep line handling checks out, and the non-UTF-8 / line-splitting edge cases are handled. Holding only on a separate issue in this file plus two hardening items.

**Blocking (not your change, but it lives here).** `vault_search` passed the query to ripgrep as a bare positional arg, so a `-`-leading query is parsed as a ripgrep *option* — `query="--pre=/bin/sh"` sets the `--pre` preprocessor and runs an arbitrary program per searched file. I've put up #51 to fix it (`-e`-guard the query). Once that merges, please rebase; your fix doesn't reintroduce it, but I'd rather land both together so the hardened argv shape is the baseline.

**Medium — symlink TOCTOU on the new context re-read.** The re-read does a lexical vault check and then separately `Path(file_path).read_bytes()`. A symlink swapped between ripgrep's match and the re-read could expand context from outside the vault. Prefer consuming ripgrep's own context events, or open with no-follow and re-validate the resolved path.

**Medium — memory.** `capture_output=True` buffers all of ripgrep's JSON before `max_results` is applied, and the fix then re-reads matched files in full. A broad query or a few huge notes can spike memory despite the result cap. Streaming the rg output / bounding the re-read would cap it.

The core fix is good — just want the rebase onto #51 and ideally the two items above.

ProfessionalSeaweedDevourer:
Rebased onto `main` and addressed the review items.

**Rebase (Blocking 1).** Now that #38 and #51 have landed, this is rebased onto `main` and carries only the `vault_search` change (no more of #38's serialization/fixtures). The argv-block conflict resolved by keeping #51's `-e` guard and dropping the now-unused `--context` flag: this PR assembles context by re-reading, so rg no longer needs to emit context events. #51's RCE regression test and the context tests both pass against the real `rg` binary.

**Symlink TOCTOU (Medium).** The context re-read now opens each path component below the vault root with `O_NOFOLLOW` relative to its parent's directory fd (`openat`), so a symlink anywhere in the path (final or intermediate, including one swapped in concurrently between rg's scan and the re-read) fails the open atomically instead of being resolved by a racy path lookup. The trusted, configured root is still opened normally, since `VAULT_PATH` may itself be a symlink. Where `openat`/`dir_fd` is unavailable (e.g. Windows) it falls back to a single `O_NOFOLLOW` open guarded by a resolved-path containment check.

I checked the race is actually closed, not just narrowed: against a plain realpath-then-open, a concurrent parent-dir swap leaked outside-vault content within a handful of attempts; against the `openat` walk, 2358 attempts across three concurrent swapping threads leaked nothing.

**Memory (Medium).** The re-read is capped at `config.MAX_SEARCH_REREAD_BYTES` (= `MAX_CONTENT_SIZE`, 1 MB); a matched file over the cap degrades to rg's matched line, so a few huge notes or a broad query can't spike memory despite the result cap. I took the "bound the re-read" half of your suggestion rather than the streaming half, to keep this PR self-contained and to avoid reworking #51's subprocess-based RCE test. Happy to do streaming as a separate focused PR if you want the broad-query buffering bounded too.

Folded in while here: a stale rg `line_number` past the file's current length (a content race where the file shrank between scan and re-read) now degrades to the matched line instead of slicing to an empty snippet.

**Tests.** Full suite green (202), with new coverage for the final-component symlink, symlinked parent, oversized file, stale `line_number`, the no-`dir_fd` fallback branch, and a deeply nested file. The symlink / size / stale / fallback tests fake the rg match event, so they run without ripgrep on PATH as well.

**Out of scope, separate issue to follow: hardlink.** A hardlink created inside the vault to a file outside it leaks the outside content here too, but this is pre-existing and vault-wide: `vault.read_file` leaks identically (the `realpath` of a hardlink stays in-vault, and `O_NOFOLLOW` doesn't apply to hardlinks). Symlinks are refused in both paths. Since it isn't introduced by this change and a real fix is a vault-wide decision (reject `st_nlink > 1`, rely on `protected_hardlinks`, or scope the mount), I've left it out of this PR and will open a separate issue with a repro.

jimprosser:
Thanks for this — the `context_lines` parity fix is right and the argv handling is correct (query passed as `-e <query>`, and the rg-dependent tests skip cleanly when ripgrep is absent). Before it can land it needs the re-read path brought fully up to the project's path-safety bar. A second review pass (mine + Codex) surfaced these:

**Path safety (blocking):**
1. The context re-read uses `O_NOFOLLOW` per component, but it doesn't run the target through `resolve_vault_path` and doesn't reject `..`. Reject `..` segments and re-validate the *resolved* re-read target (keep the component-wise `O_NOFOLLOW`), and fail closed on platforms where that can't be done safely.
2. Hardlinks: this is the one I want to flag clearly. Because `context_lines` was previously a no-op, the surrounding lines were never returned; now they are — so for an in-vault hardlink pointing outside the vault, this PR widens the leak from "the matched line" to "the matched line plus its context." I've decided the hardlink policy on #53: **refuse to read any file with `st_nlink > 1`, fail closed**, documented as "legitimate in-vault hardlinks aren't supported." I'd like that guard to land with this PR so the search path is covered end-to-end.
3. The same safe-read policy needs to cover the other read paths in `vault_search`, not just the rg context re-read — `_get_frontmatter_excerpt` still uses a plain `read_text` that follows links, and the Python fallback should go through the same guard.

**Tests (blocking):**
4. The symlink abuse tests exercise `_search_ripgrep`, not the full `vault_search` tool path. Add full-tool abuse tests: symlink race, hardlink leakage (asserting the `st_nlink > 1` file is refused), and the rg-absent Python fallback.

The hardlink guard (#53) is the gating piece — happy to coordinate on it landing here vs. as a small precursor PR, your call. Everything else is mechanical once that's in.

ProfessionalSeaweedDevourer:
Addressed the four blocking items (pushed `b32955e`).

**1. Parent-traversal + resolved-target re-validation.** The context re-read target is now gated through `resolve_vault_path` (rejects `..`, dotfiles, null bytes, and out-of-vault resolution) before the openat walk; `..` mattered because O_NOFOLLOW does not stop a `..` component (a real directory entry that walks up out of the vault). After the O_NOFOLLOW open, the fd is re-validated as inside the vault via the OS fd->path facility (`F_GETPATH` on macOS, `/proc/self/fd` on Linux), as defense in depth behind the openat walk.

**2. #53 hardlink guard, here.** Every read path refuses a file with `st_nlink > 1`, fail closed, so an in-vault hardlink to an outside file can no longer leak (or, via context, widen a leak). Legitimate in-vault hardlinks are unsupported as a result.

**3. Same safe-read across paths.** All three search read paths now go through one `_safe_read_vault_bytes` (gate + openat O_NOFOLLOW + fd re-validation + `st_nlink` + optional size cap): the ripgrep context re-read, the Python search fallback (which previously followed symlinks outright), and `_get_frontmatter_excerpt`. `read_file` gets the same `st_nlink` guard. The context re-read keeps its memory size cap; the Python fallback intentionally has none, since that read *is* the search and a cap would silently drop matches in large files.

**4. Full-tool abuse tests.** parent-traversal refused; hardlink refused on the re-read, the Python fallback, the frontmatter excerpt and `read_file`; symlink not followed by the Python fallback (rg-absent path); fd re-validation accepts in-vault / rejects outside fds; plus regression guards that normal single-link files still read. The existing symlink-swap and oversized-file tests are unchanged.

Full suite green (215 passed). #53 lands with this PR as requested.

jimprosser:
Thanks for this — the context_lines diagnosis is right and the encoding/line-numbering care (the `_split_lines` vs `splitlines()` desync, CRLF, U+2028) is exactly the subtle part most people would miss.

But before this merges I need to flag that **#53 isn't actually closed on the ripgrep path** — which is the default whenever `rg` is installed — and a test is masking it.

## The hardlink leak survives on the rg backend

A `vault_search` that matches inside an in-vault hardlink-to-an-outside file still returns the outside content:

```python
import os, tempfile
from pathlib import Path
import obsidian_vault_mcp.config as config
import obsidian_vault_mcp.tools.search as search_mod

root = tempfile.mkdtemp()
outside = Path(root, "outside"); outside.mkdir()
vault = Path(root, "vault"); vault.mkdir()
secret = outside / "secret.txt"
secret.write_text("line before\nSECRET hunter2 token\nline after\n")
os.link(secret, vault / "evil.md")          # in-vault hardlink, st_nlink == 2

config.VAULT_PATH = vault
print(search_mod._search_ripgrep("SECRET", vault, file_pattern="*.md",
                                 max_results=10, context_lines=2))
# -> [{'path': 'evil.md', 'line_number': 2, 'match_context': 'SECRET hunter2 token'}]
```

(With `rg` absent, the Python fallback correctly returns `[]` — so this is a backend asymmetry, and the leaky one is the default.)

The chain: ripgrep reads `evil.md` (a hardlink is indistinguishable from a normal file — unlike a symlink, which rg skips) and emits a match event whose `lines.text` is the outside content. The new guard *does* fire — it even logs `refusing hardlinked file` — but the loop then **degrades to ripgrep's matched line** (`_rg_match_line`) and appends the match anyway. The guard suppresses the surrounding *context*, not the matched line, and the matched line is the leak.

Root cause: `_read_vault_file_lines` returns `None` for both a benign re-read failure (file shrank mid-scan) and a **security refusal** (hardlink / resolves-outside). The rg loop treats both as "degrade to rg's line." For a security refusal the match has to be **dropped**, the way the Python fallback already drops it — not degraded.

## The test that should catch this is mocking past it

`test_ripgrep_reread_refuses_hardlinked_match` monkeypatches `subprocess.run` to return a hardcoded `"decoy match"` line instead of letting real `rg` read the real secret, then asserts the decoy comes back clean. That only proves the *context* was suppressed — it can't prove the matched line is safe, because the secret was never in the faked output. The other `_rg_match_json`-based hardening tests have the same blind spot. A regression test here needs real `rg` reading a real hardlinked file and asserting zero results (skip when `rg` is absent, like the other `@requires_rg` cases).

## I'd like to split this

The more I look, the more I think the context fix and the #53 fix are two separate changes wearing one hat, and bundling them is what let the leak hide:

1. **The `context_lines` fix on its own.** The minimal version is to consume ripgrep's own `context` events — rg was already emitting them under `--context`, the original code just dropped them on the floor. That fixes the documented bug without a second read at all, which deletes the entire re-read path (`_open_in_vault_nofollow`, `_fd_resolves_inside_vault`, the size cap) and the attack surface that came with it. On a tunnel-exposed personal-vault server I'd rather not add that much new filesystem machinery to fix a context bug.

2. **The #53 fix on its own.** The `read_file` `st_nlink > 1` guard is correct and can land as-is. The search side needs the rg backend to *drop* matches whose file fails the safety check (hardlink/outside), matching the Python fallback — plus the real-`rg` test above.

Two smaller things I noticed while in here, lower priority and both local-actor-only, but worth a note if the re-read survives in any form:

- **FIFO hang.** The re-read opens `O_RDONLY | O_NOFOLLOW` with no `O_NONBLOCK` and no `S_ISREG` check. A matched file swapped for a FIFO before the re-read makes the open block indefinitely — the rg subprocess has a 30s timeout, this read has none.
- **Size cap is pre-read only.** `MAX_SEARCH_REREAD_BYTES` is checked against `st_size` once; the read loop has no cumulative cap, so a file growing during the read isn't bounded.

None of this is a knock on the work — the diagnosis and the line-counting handling are genuinely good. It's the rg degrade path and the mocked test that need to change before I can call #53 closed. Happy to take it as two PRs in whatever order suits you.

jimprosser:
Checking in after the June round. I re-ran the hardlink repro against your current head with real ripgrep and the rg backend still returns the matched line from the hardlinked file, so #53 is still open on the default path. The split I proposed stands: the `context_lines` fix on its own by consuming rg's `context` events, and the #53 guard as a separate PR where the rg backend drops the match rather than degrading to it, with a real-rg regression test. If I don't hear back in a couple of weeks I'll close this and lift the `read_file` `st_nlink` guard out as a small change with credit to you.

jimprosser:
Following up on my September check-in rather than leaving this hanging: I've landed the #53 half on its own in #79, with credit to you on the issue. Your diagnosis drove it, including the call that the hardlink case is vault-wide rather than a search bug.

That's the split I proposed in June, done from this end so the security hole isn't waiting on a round-trip — it had been open three months. What landed is the `st_nlink > 1` refusal across all four read paths, with the ripgrep backend dropping a refused match instead of degrading to the line ripgrep already returned. The helper raises on a security refusal now rather than returning `None`, which is what let the two cases get conflated at the call site.

Two notes from doing it, in case they're useful:

The test bar turned out to be the hard part, and it's worth saying plainly because it isn't a knock on your work — it's a trap in this repo's environment. On my machine `rg` is a shell function rather than a binary, so `shutil.which("rg")` returns `None` from Python and every ripgrep-dependent test silently skips. A green suite reporting "1 skipped" looks identical to a pass. That is very likely why the hardlink case read as closed for three months on both our ends. The new tests assert real ripgrep actually emits the secret *before* asserting the guard suppresses it, so they can't pass the way the mocked-decoy version did.

`main` was also leaking more than either of us had written down: the Python fallback returned the matched line plus context, because the guard that made it return `[]` only ever existed on your branch.

**This PR is still yours and I'd like it.** What's left is the original `context_lines` fix on its own — ripgrep already emits `context` events under `--context` and the old code dropped them, so consuming those fixes the documented bug with no second read at all, and none of the re-read machinery. `main` has moved, so it'll want a rebase. No rush and no clock on it this time.

Issue #1, opened 2026-03-21
Claude.ai cannot connect: MCP endpoint at /mcp but Claude.ai posts to /


ben-bl:
## Bug

Claude.ai fails to connect to the server after successful OAuth authentication. The OAuth flow completes (authorization code issued, token exchanged) but all subsequent MCP requests return 404.

## Root Cause

The MCP library defaults `streamable_http_path` to `/mcp`, but Claude.ai always posts to `/` after authentication. Claude.ai does not use the RFC 9728 `/.well-known/oauth-protected-resource` endpoint to discover the MCP path — I confirmed this by implementing the endpoint and watching Claude.ai fetch it successfully, then still post to `/`.

### Request log showing the issue

```
POST /oauth/token HTTP/1.1  → 200 OK        # auth succeeds
POST / HTTP/1.1             → 404 Not Found  # MCP request fails
GET / HTTP/1.1              → 404 Not Found  # retry fails
```

## Steps to Reproduce

1. Follow the README setup (clone, configure, run)
2. Connect via Claude.ai → Settings → Integrations → Add Integration
3. OAuth flow completes successfully
4. Claude.ai shows "Authorization with the MCP server failed"

## Suggested Fix

Add `streamable_http_path="/"` to the `FastMCP` constructor in `server.py`:

```python
mcp = FastMCP(
    "obsidian_web_mcp",
    stateless_http=True,
    json_response=True,
    streamable_http_path="/",  # Claude.ai expects MCP at root
    lifespan=lifespan,
    ...
)
```

This is a one-line change. The OAuth routes are mounted with higher priority (inserted at position 0 in the route list), so they still match before the MCP catch-all at `/`.

Additionally, implementing RFC 9728 `/.well-known/oauth-protected-resource` would help other MCP clients discover the endpoint properly, even though Claude.ai currently ignores it.

## Environment

- obsidian-web-mcp: initial release (commit 5eb1e88)
- mcp library: 1.26.0
- Claude.ai web client (March 2026)
- macOS, behind Cloudflare Tunnel

PR #40, opened 2026-06-08
vault_edit: expose edit-field schema, expand aliases, diagnose mismatches

ProfessionalSeaweedDevourer:
> **Stacked on #38 — assumes it is merged first.** This branch builds on #38 (non-ASCII serialization): the `vault_edit` diagnostics return through `serialization.dumps`, which #38 introduces in `write.py`. Until #38 merges, the diff below also shows its commits; it narrows to just the diagnostics once #38 lands. #38's `write.py` change is the `json.dumps -> dumps` swap, not the `vault_edit` logic edited here, so the two do not conflict.
>
> _(Previously this branch also carried #39 (ripgrep context_lines); that is independent of these diagnostics and has been dropped from here — review it as its own PR #39.)_

## What

Three follow-ups to a `vault_edit` usage-friction log, each implemented TDD
(test failing first, then minimal code). Adds `tests/test_vault_edit_friction.py`
(10 cases); full suite green (**200 passed**), ruff clean on touched files.

### P0 — expose the per-edit field schema + widen the alias set

`vault_edit` was registered with `edits: list[dict]`, so FastMCP advertised
each edit as an opaque `{"type": "object", "additionalProperties": true}`. A
caller could not discover the `old_text` / `new_text` field names from the
schema and had to guess on the first call (the original friction).

- Type the parameter as `edits: list[VaultEditOperationInput]`. FastMCP now
  emits `edits.items` as `{"$ref": "#/$defs/VaultEditOperationInput"}` with
  `old_text` / `new_text` properties, so the fields are self-describing.
- Accept `old` / `new` aliases in addition to `old_str` / `new_str` (the
  shorthand that gets mis-guessed most in practice).
- Consolidate alias handling into one `normalize_edit_aliases` in `models.py`,
  shared by the model (the MCP-schema path) and the write tool (the
  direct-call path), so the two alias sets cannot drift apart. Supplying a
  canonical key together with one of its aliases — or two aliases — is rejected
  with the offending keys named.
- One line added to the tool description spelling out the `{old_text, new_text}`
  shape, for clients that do not read the schema.

### P1 — near-miss diagnostics on a zero-match edit

When `old_text` matches zero times, the error now carries a `near_miss` object
(`line_number`, `line`, `similarity`) pointing at the most similar line in the
file, so a one-character typo is obvious without re-reading the whole document.
Candidate search is line-scoped, keeping cost linear in file size; unrelated
input yields a low similarity rather than a misleading "match".

### P2 — dry-run reports all match counts against the original

`dry_run=True` no longer fails fast at the first non-unique match. It counts
each edit's `old_text` against the **original** document and returns every
result in `match_counts` (`0` / `1` / many), so one preview surfaces all the
mismatches at once. When every edit matches exactly once the sequential diff
preview and `edits_applied` are still included (existing dry-run contract
preserved). Real application (`dry_run=False`) keeps the sequential fail-fast
safety net unchanged.

## Files

| File | Change |
|------|--------|
| `models.py` | `normalize_edit_aliases` shared helper; `old`/`new` aliases; multi-alias conflict detection |
| `server.py` | `edits: list[VaultEditOperationInput]`; description line |
| `tools/write.py` | delegate alias normalization to the shared helper; `_find_near_miss`; `_dry_run_report` |
| `tests/test_vault_edit_friction.py` | 10 new tests (P0/P1/P2) |

## Notes

- The pre-existing `_normalize_edit_aliases` in `write.py` looked dead but is
  live on the direct-call path (the existing tests call `vault_edit` directly
  with alias keys), so it was kept as a thin delegator to the shared helper
  rather than deleted.
- `vault_batch_frontmatter_update` has the same `updates: list[dict]`
  schema-opacity as P0 addressed here; left for a separate follow-up.

### DIFF
--- a/src/obsidian_vault_mcp/models.py
+++ b/src/obsidian_vault_mcp/models.py
@@ -53,6 +53,37 @@ class VaultWriteInput(BaseModel):
     )
 
 
+# Aliases accepted for each canonical edit field. old_str/new_str mirror the
+# str_replace_editor tool; old/new are the shorthands models reach for most.
+_EDIT_FIELD_ALIASES: dict[str, tuple[str, ...]] = {
+    "old_text": ("old_str", "old"),
+    "new_text": ("new_str", "new"),
+}
+
+
+def normalize_edit_aliases(data):
+    """Map edit-field aliases onto their canonical old_text/new_text keys.
+
+    Shared by the pydantic model (the MCP schema path) and the write tool (the
+    direct-call path) so both accept the same alias set. Raises ValueError,
+    naming the offending keys, if a canonical field and one of its aliases (or
+    two aliases) are supplied together.
+    """
+    if not isinstance(data, dict):
+        return data
+
+    normalized = dict(data)
+    for canonical, aliases in _EDIT_FIELD_ALIASES.items():
+        present = [key for key in (canonical, *aliases) if key in normalized]
+        if len(present) > 1:
+            joined = ", ".join(f"'{key}'" for key in present)
+            raise ValueError(f"Use only one of {joined}, not several")
+        if present and present[0] != canonical:
+            normalized[canonical] = normalized.pop(present[0])
+
+    return normalized
+
+
 class VaultEditOperationInput(BaseModel):
     """Replace one exact text fragment inside a vault file."""
 
@@ -61,17 +92,7 @@ class VaultEditOperationInput(BaseModel):
     @model_validator(mode="before")
     @classmethod
     def normalize_str_replace_aliases(cls, data):
-        if not isinstance(data, dict):
-            return data
-
-        normalized = dict(data)
-        for canonical, alias in (("old_text", "old_str"), ("new_text", "new_str")):
-            if canonical in normalized and alias in normalized:
-                raise ValueError(f"Use either '{canonical}' or '{alias}', not both")
-            if alias in normalized:
-                normalized[canonical] = normalized.pop(alias)
-
-        return normalized
+        return normalize_edit_aliases(data)
 
     old_text: str = Field(
         ...,
--- a/src/obsidian_vault_mcp/models.py
+++ b/src/obsidian_vault_mcp/models.py
@@ -1,6 +1,6 @@
 """Pydantic input models for obsidian-vault-mcp tool endpoints."""
 
-from typing import Literal
+from typing import Any, Literal
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
@@ -61,7 +61,7 @@ _EDIT_FIELD_ALIASES: dict[str, tuple[str, ...]] = {
 }
 
 
-def normalize_edit_aliases(data):
+def normalize_edit_aliases(data: Any) -> Any:
     """Map edit-field aliases onto their canonical old_text/new_text keys.
 
     Shared by the pydantic model (the MCP schema path) and the write tool (the
--- a/src/obsidian_vault_mcp/server.py
+++ b/src/obsidian_vault_mcp/server.py
@@ -129,9 +129,9 @@ def vault_write(path: str, content: str, create_dirs: bool = True, merge_frontma
     description=(
         "Patch an existing vault file with exact text replacements. Use this for token-efficient partial edits "
         "when only small fragments change; supports dry-run diff previews and avoids resending the full file. "
-        "Each edit is an object {old_text, new_text}; old_text must match exactly once. dry_run counts every "
-        "edit's old_text against the ORIGINAL file, so a chained sequence (one edit's new_text becomes a later "
-        "edit's old_text) can preview as all-unique yet still fail on apply, which runs the edits in order."
+        "Each edit is an object {old_text, new_text}; old_text must match exactly once. Edits apply in order, "
+        "and dry_run simulates that same in-order apply (each old_text is matched against the running document "
+        "the earlier edits produce), so its diff and match counts predict the real apply, including chained edits."
     ),
     annotations={"readOnlyHint": False, "destructiveHint": True, "idempotentHint": False, "openWorldHint": False},
 )
--- a/src/obsidian_vault_mcp/server.py
+++ b/src/obsidian_vault_mcp/server.py
@@ -79,6 +79,7 @@ from .models import (
     VaultReadInput,
     VaultWriteInput,
     VaultEditInput,
+    VaultEditOperationInput,
     VaultAppendInput,
     VaultBatchReadInput,
     VaultBatchFrontmatterUpdateInput,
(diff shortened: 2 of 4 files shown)


jimprosser:
The alias work itself is clean — canonical+alias conflicts are rejected, `extra="forbid"` blocks unknown keys on the MCP path, the empty-`old_text` guard is a real safety fix, and the replace logic is unchanged so edits can't hit the wrong field or note. No auth or path-traversal issues, and the diagnostics only echo a line from the already-authorized target file. Two blockers.

**Blocking 1 — this PR is stacked on #38.** The diff currently carries #38's serialization change and its UDHR fixtures/tests (`main` has neither), which is ~700 of the lines here. Please rebase onto `main` after #38 lands so this reviews as just the `vault_edit` change (`models.py`, `server.py`, `tools/write.py`, `tests/test_vault_edit_friction.py`). As-is, merging this first would pull #38 in unreviewed.

**Blocking 2 — the near-miss diagnostic is an authenticated CPU DoS (`tools/write.py`, `_find_near_miss`).** On a zero-match edit it runs `difflib.SequenceMatcher` per line against the caller-supplied `old_text`. A ~1 MB high-entropy `old_text` that doesn't match pins the process (a bounded probe with 1 MB random text vs 1000 short lines didn't return in minutes), and `old_text` is bounded only by the ~1 MB/edit, 20-edit limits. Cap or skip near-miss above a small `old_text` size, or use a cheaper bounded scorer.

**Nit.** The dry-run match counts are computed independently against the original document, so chained edits (one edit's `new_text` feeds another's `old_text`) can make dry-run report "all unique/safe" for a sequence that fail-fasts on apply. It's documented in the docstring — worth surfacing in the user-facing tool description too, since dry-run is exactly when someone trusts the preview.

ProfessionalSeaweedDevourer:
Rebased onto `main` and addressed the review.

**Rebase (Blocking 1).** With #38 and #51 merged, this is rebased so the diff is just the `vault_edit` change (`models.py`, `server.py`, `tools/write.py`, `tests/test_vault_edit_friction.py`); the #38 serialization change and UDHR fixtures are dropped.

**Near-miss CPU DoS (Blocking 2).** `_find_near_miss` ran one `difflib.SequenceMatcher` pass per line against the caller-supplied `old_text`. Two bounds now:

- Skip the scan for an `old_text` larger than 1 KB: not a plausible single-line near-miss, and the input you reproduced (~1 MB high-entropy `old_text` pinned the process).
- A cumulative `len(old_text) * len(line)` work budget, since even a capped `old_text` against a large many-line file still ran away (~15 s for a 1 MB / 25k-line file, and the edited file's size isn't otherwise bounded). The loop stops once the budget is spent.

That proxy underestimates difflib's constant factor for low-alphabet, autojunk-disabled (sub-200-char) lines, so the budget is sized to keep even an adversarially crafted file under ~1 s (measured 0.8 s for a 2-char alphabet with 199-char lines), versus minutes unbounded. Typo-scale `old_text` still scans thousands of lines, so normal notes keep full near-miss coverage.

**Dry-run caveat (Nit).** The tool description now states that `dry_run` counts each edit's `old_text` against the original file, so a chained sequence (one edit's `new_text` becomes a later edit's `old_text`) can preview as all-unique yet still fail on apply, which runs the edits in order.

Tests: full suite green; new coverage for an oversized `old_text` (skipped), a large high-entropy `old_text` (returns promptly), a capped `old_text` over a 25k-line file, and an adversarial low-alphabet file, all staying bounded.

jimprosser:
Thanks — the edit-field schema, alias expansion, and near-miss diagnostics are a real ergonomics improvement, and I appreciate that you bounded the near-miss CPU and kept the empty-`old_text` rejection and the atomic all-or-nothing apply intact. A second review pass (mine + Codex) leaves two things before it can land:

**Dry-run / apply parity (blocking):** `_dry_run_report` counts every `old_text` against the *original* document, but the real apply runs the edits in *sequence*. So a chained edit set — where one edit's `new_text` becomes a later edit's `old_text` — can preview as all-unique yet fail on apply (and you've documented exactly this caveat in the tool description, which is the tell). A dry run needs to preview the write that would actually pass: run the same sequential exact-once validation as apply when generating the diff. Keep the aggregate per-edit match counts as separate diagnostics if they're useful, but the diff itself shouldn't green-light a sequence the apply path will reject.

**Wiring tests (blocking):** the alias/conflict/empty-`old_text` behavior is tested through the model and helper, and the schema through `mcp.list_tools`, but not through the registered `server.vault_edit` wrapper. Add wrapper-level tests for `old`/`new` aliases, an alias conflict, empty `old_text`, and "file unchanged on failure" so the tool seam is covered, not just the helpers.

Add a negative dry-run test for the chained-duplicate case as part of the first item. Once those are in I'll merge.

ProfessionalSeaweedDevourer:
Addressed both blockers from the last pass (pushed `94168ce`).

**1. dry-run/apply parity.** `_dry_run_report` now simulates the in-order apply instead of counting each `old_text` against the original document: each count is measured against the running document the preceding edits produce, a unique match is folded into it before the next edit, and the diff is emitted only when every edit matches exactly once. A chained set whose first edit feeds or duplicates a later edit's `old_text` is now predicted exactly as it applies. This also removed the old re-apply loop, which silently no-op'd edits that no longer matched in the mutated preview while still reporting `edits_applied` as the full count.

Negative tests added (each asserts dry-run and the real apply agree):
- chained duplicate: edit 0's `new_text` introduces a second occurrence of edit 1's `old_text` -> dry-run reports count 2, `edits_applied` 0, empty diff; the real apply fails.
- sequential consume: edit 0 removes edit 1's `old_text` -> dry-run reports count 0; the real apply fails.

**2. wrapper-level tests through the registered `server.vault_edit`:**
- old/new alias normalization applies the edit;
- alias conflict (`old_text` + `old`) -> rejected by the input model (`ValidationError`) before any write;
- empty `old_text` -> rejected by the model (`min_length=1`) before the write tool runs (defense-in-depth, distinct from the direct call's JSON error);
- a well-formed non-matching edit -> JSON error, file unchanged;
- mid-sequence failure -> all-or-nothing: the earlier matching edit is not partially written.

Full suite green (205 passed). Tool description and `_dry_run_report` docstring updated to state the parity. Ready for another look.

jimprosser:
This is clean work — and a notable catch beyond the stated scope: the old `dry_run` counted matches against the *original* document, so a chained edit set could preview as applicable and then fail on the real apply. Simulating the sequential apply (folding each match into a running document) with parity tests against the real path is exactly right, and the near-miss CPU guard — size cap plus cumulative work budget, with wall-clock regression tests — is a real fix to an authenticated DoS, not a token gesture. Match-exactly-once, the no-partial-write guarantee, and empty-`old_text` rejection all hold and are tested at the registered-tool level.

A few small things before I merge — only the first really matters:

1. **`dry_run` doesn't predict a `MAX_CONTENT_SIZE` failure.** The new tool description says dry_run "predicts the real apply," but the 1 MB cap is only enforced in `write_file_atomic`. Replace a fragment in a 900 KB note with a 200 KB string and dry_run returns a diff with `edits_applied: 1`, while the real apply raises and writes nothing. No data-loss bug — but the "predicts the real apply" wording is inaccurate for that case. Either check the predicted size in `_dry_run_report` and report it as not-applicable, or soften the description.

2. **`near_miss` is best-effort on large files.** The work budget can stop the scan before a later, more-similar line, so the returned line is "closest found within budget," not guaranteed the global best. Good trade for the DoS guard — just worth wording it as best-effort so it's not read as a guarantee.

3. **PR description drift.** The description still says dry_run counts against the original document; the code does the better thing (sequential simulation). Worth updating so the description matches what shipped.

Optional, not blocking: strengthen the schema test to assert `items.properties.old_text/new_text` and `additionalProperties: false` rather than substring presence, and add alias-conflict cases for the `new_text`/`new`/`new_str` group and canonical-plus-both-aliases.

Fix #1 (or reword it) and I'm happy to merge.

jimprosser:
Picking this up: you addressed both rounds and the only open item was the `MAX_CONTENT_SIZE` prediction in `dry_run`. I've rebased onto main (two adjacent-insertion conflicts, suite green at 398) and added that check with a parity test as #78, with your four commits preserved as authored. #78 closes this one when it merges. Thanks for the sequential-simulation catch; that was the best part of this.

Issue #4, opened 2026-03-21
Security: No redirect_uri validation + missing client_id verification in token exchange


ben-bl:
## Security Issues

Two related OAuth implementation gaps that together enable authorization code interception.

### 1. No redirect_uri validation in /oauth/authorize

The authorization endpoint accepts any `redirect_uri` without validating it against URIs provided during client registration:

```python
# oauth.py, oauth_authorize()
redirect_uri = request.query_params.get("redirect_uri", "")
# ... no validation against registered URIs ...
return RedirectResponse(url=f"{redirect_uri}{separator}{urlencode(params)}")
```

An attacker can supply `redirect_uri=https://attacker.com/callback` and the authorization code will be redirected there.

### 2. Auth code exchange doesn't verify client_id

Per RFC 6749 §4.1.3, the token endpoint MUST verify that the authorization code was issued to the authenticated client. The current implementation stores `client_id` in the auth code data but never checks it during exchange:

```python
# oauth.py, _handle_authorization_code()
code_data = _auth_codes.pop(code)

# Verify redirect_uri matches ✓
# Verify PKCE ✓
# Verify client_id ✗  (not checked)
# Verify client_secret ✗  (not checked)
```

### Combined Impact

An attacker who obtains an authorization code (via the redirect_uri issue or other means) can exchange it from any client without proving they are the intended recipient.

## Suggested Fixes

**For redirect_uri:** Store allowed redirect URIs during registration, validate the authorize request's `redirect_uri` against the stored list using exact string matching. For the default `vault-mcp-client`, hardcode the allowed Claude.ai callback URL (`https://claude.ai/api/mcp/auth_callback`).

**For client_id verification:** Add to `_handle_authorization_code`:

```python
if client_id != code_data["client_id"]:
    return JSONResponse(
        {"error": "invalid_grant", "error_description": "client_id mismatch"},
        status_code=400,
    )
```

Also verify `client_secret` during the authorization code exchange, not just during the client_credentials grant.

## Severity

High. These are well-known OAuth vulnerability classes (OWASP A07:2021).

mleitnercom:
Fix in fork: /oauth/authorize now validates client_id + redirect_uri; token exchange verifies client_id and client_secret before issuing tokens. Commit: https://github.com/mleitnercom/obsidian-web-mcp/commit/bba288308b0502d063ff4535444a5996a9ff327d. Happy to open a PR if you want.

jjsmackay:
[...]

Issue #3, opened 2026-03-21
Security: /oauth/register leaks client secret to unauthenticated callers


ben-bl:
## Security Issue

The dynamic client registration endpoint (`POST /oauth/register`) is unauthenticated and returns the real `VAULT_OAUTH_CLIENT_SECRET` in its response:

```python
# oauth.py, oauth_register()
return JSONResponse({
    "client_id": client_id,
    "client_secret": config.VAULT_OAUTH_CLIENT_SECRET,  # exposed to any caller
    ...
}, status_code=201)
```

Since `/oauth/register` is in the `_AUTH_EXEMPT_PATHS` set, any HTTP client can retrieve the secret without authentication.

## Attack Chain

Combined with the auto-approve authorization flow, an attacker can:

1. `POST /oauth/register` → receives the `client_secret`
2. `GET /oauth/authorize?redirect_uri=https://attacker.com/callback&...` → receives authorization code redirected to their server (see: no redirect_uri validation — separate issue)
3. `POST /oauth/token` with the code → receives the bearer token
4. Full read/write access to the vault

## Suggested Fix

Options (in order of preference):

1. **Don't return the secret in registration responses.** Require clients to present the secret as a pre-shared value during the token exchange, rather than distributing it via registration.

2. **Remove dynamic registration entirely.** For a single-user server, pre-configured credentials (as documented in the README) are sufficient. Claude.ai works fine with static `vault-mcp-client` / secret.

3. **Rate-limit or require authentication on `/oauth/register`.** This is harder since it creates a bootstrap problem.

## Severity

Critical at the application layer. Mitigated if the server is behind Cloudflare Access, but the README's deployment model (Cloudflare Tunnel without Access) leaves this exposed.

mleitnercom:
Fix in fork: /oauth/register no longer returns the server’s shared secret; registration returns a per‑client secret and token exchange validates it. Commit: https://github.com/mleitnercom/obsidian-web-mcp/commit/bba288308b0502d063ff4535444a5996a9ff327d. Happy to open a PR if you want.

jjsmackay:
[...]

PR #42, opened 2026-06-09
Preserve YAML frontmatter formatting on merge writes

jjsmackay:
**Problem:** `vault_write(merge_frontmatter=True)` rewrites YAML formatting on every merge. PyYAML normalises quotes, flattens block lists, turns `yes`/`no` into `true`/`false`, drops comments, reorders keys. Obsidian Linter / yamllint users lose their formatting on each write.

**Fix:** route the merge through a `ruamel.yaml` round-trip wrapper (`frontmatter_io.py`). It mutates the existing `CommentedMap` in place, so untouched keys keep their formatting and new keys append.

**Scope:**
- Only `merge_frontmatter=True` changes; `=False` writes stay byte-identical.
- Read/index paths and `vault_batch_frontmatter_update` keep python-frontmatter (formatting not observed there).

**Tests:** 84 pass, 12 new: module round-trip plus end-to-end via `vault_write`. No existing test asserted the normalised output.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ requires-python = ">=3.12"
 dependencies = [
     "mcp[cli]>=1.9.0",
     "python-frontmatter>=1.1.0",
+    "ruamel.yaml>=0.18",
     "pydantic>=2.0",
     "watchdog>=4.0.0",
 ]
--- a/src/obsidian_vault_mcp/frontmatter_io.py
+++ b/src/obsidian_vault_mcp/frontmatter_io.py
@@ -1,0 +1,76 @@
+"""YAML frontmatter I/O that preserves formatting across round-trips.
+
+Uses ruamel.yaml in round-trip mode so quote style, comments, block/flow
+style, boolean forms (yes/no vs true/false), and key order survive a
+load-then-dump cycle. PyYAML (via python-frontmatter) normalizes all of
+these, which rewrites users' carefully-formatted frontmatter on every
+update.
+"""
+
+from __future__ import annotations
+
+import io
+import logging
+import re
+
+from ruamel.yaml import YAML
+from ruamel.yaml.error import YAMLError
+
+logger = logging.getLogger(__name__)
+
+_FRONTMATTER_RE = re.compile(
+    r"\A---[ \t]*\r?\n(.*?)(?:\r?\n)?---[ \t]*\r?\n?(.*)\Z",
+    re.DOTALL,
+)
+
+
+_YAML = YAML(typ="rt")
+_YAML.preserve_quotes = True
+_YAML.width = 4096
+_YAML.indent(mapping=2, sequence=4, offset=2)
+
+
+def loads(content: str) -> tuple[dict, str]:
+    """Parse a markdown file into (metadata, body).
+
+    When frontmatter is present, metadata is a ruamel.yaml CommentedMap that
+    retains the original formatting for round-trip dumping. When absent,
+    returns ({}, content).
+    """
+    match = _FRONTMATTER_RE.match(content)
+    if match is None:
+        return {}, content
+
+    raw_yaml, body = match.group(1), match.group(2)
+
+    if raw_yaml.strip() == "":
+        return {}, body
+
+    # Ensure the YAML text ends with a newline so ruamel correctly parses
+    # trailing-newline chomping on literal/folded block scalars at EOF.
+    if not raw_yaml.endswith("\n"):
+        raw_yaml += "\n"
+
+    try:
+        metadata = _YAML.load(raw_yaml)
+    except YAMLError as e:
+        logger.warning("YAML frontmatter parse failed: %s", e)
+        return {}, content
+
+    if metadata is None:
+        return {}, body
+
+    return metadata, body
+
+
+def dumps(metadata: dict | None, body: str) -> str:
+    """Serialize (metadata, body) back to a markdown file.
+
+    Empty metadata writes the body unchanged (no delimiters).
+    """
+    if not metadata:
+        return body
+
+    buf = io.StringIO()
+    _YAML.dump(metadata, buf)
+    return f"---\n{buf.getvalue()}---\n{body}"
--- a/src/obsidian_vault_mcp/frontmatter_io.py
+++ b/src/obsidian_vault_mcp/frontmatter_io.py
@@ -10,13 +10,11 @@ update.
 from __future__ import annotations
 
 import io
-import logging
 import re
+import sys
 
 from ruamel.yaml import YAML
-from ruamel.yaml.error import YAMLError
-
-logger = logging.getLogger(__name__)
+from ruamel.yaml.error import YAMLError  # noqa: F401 - re-exported for callers
 
 _FRONTMATTER_RE = re.compile(
     r"\A---[ \t]*\r?\n(.*?)(?:\r?\n)?---[ \t]*\r?\n?(.*)\Z",
@@ -24,10 +22,21 @@ _FRONTMATTER_RE = re.compile(
 )
 
 
-_YAML = YAML(typ="rt")
-_YAML.preserve_quotes = True
-_YAML.width = 4096
-_YAML.indent(mapping=2, sequence=4, offset=2)
+def _make_yaml() -> YAML:
+    """Build a fresh round-trip handler.
+
+    ruamel.yaml's YAML object holds mutable parser/emitter state and is not
+    reentrant, so a module-level singleton corrupts under concurrent use
+    (FastMCP runs sync tools in a threadpool). Construct one per call --
+    cheap for human-triggered writes.
+    """
+    yaml = YAML(typ="rt")
+    yaml.preserve_quotes = True
+    # Disable line wrapping: any finite width re-folds long scalars (URLs,
+    # descriptions) on dump, which is exactly the churn this module avoids.
+    yaml.width = sys.maxsize
+    yaml.indent(mapping=2, sequence=4, offset=2)
+    return yaml
 
 
 def loads(content: str) -> tuple[dict, str]:
@@ -35,7 +44,9 @@ def loads(content: str) -> tuple[dict, str]:
 
     When frontmatter is present, metadata is a ruamel.yaml CommentedMap that
     retains the original formatting for round-trip dumping. When absent,
-    returns ({}, content).
+    returns ({}, content). Raises YAMLError when delimiters are present but
+    the enclosed YAML is invalid -- the caller decides how to handle it,
+    rather than silently conflating "no frontmatter" with "broken frontmatter".
     """
     match = _FRONTMATTER_RE.match(content)
     if match is None:
@@ -51,11 +62,7 @@ def loads(content: str) -> tuple[dict, str]:
     if not raw_yaml.endswith("\n"):
         raw_yaml += "\n"
 
-    try:
-        metadata = _YAML.load(raw_yaml)
-    except YAMLError as e:
-        logger.warning("YAML frontmatter parse failed: %s", e)
-        return {}, content
+    metadata = _make_yaml().load(raw_yaml)
 
     if metadata is None:
         return {}, body
@@ -66,11 +73,19 @@ def loads(content: str) -> tuple[dict, str]:
 def dumps(metadata: dict | None, body: str) -> str:
     """Serialize (metadata, body) back to a markdown file.
 
-    Empty metadata writes the body unchanged (no delimiters).
+    Empty metadata writes the body unchanged (no delimiters). The frontmatter
+    block matches the body's line endings so a CRLF file never gains mixed
+    endings (ruamel always emits "\\n").
     """
     if not metadata:
         return body
 
     buf = io.StringIO()
-    _YAML.dump(metadata, buf)
-    return f"---\n{buf.getvalue()}---\n{body}"
+    _make_yaml().dump(metadata, buf)
+    yaml_text = buf.getvalue()
+
+    newline = "\r\n" if "\r\n" in body else "\n"
+    if newline != "\n":
+        yaml_text = yaml_text.replace("\n", newline)
+
+    return f"---{newline}{yaml_text}---{newline}{body}"
--- a/src/obsidian_vault_mcp/frontmatter_io.py
+++ b/src/obsidian_vault_mcp/frontmatter_io.py
@@ -10,16 +10,27 @@ update.
 from __future__ import annotations
 
 import io
-import re
 import sys
 
 from ruamel.yaml import YAML
-from ruamel.yaml.error import YAMLError  # noqa: F401 - re-exported for callers
+from ruamel.yaml.error import YAMLError  # re-exported for callers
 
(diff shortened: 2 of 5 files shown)


ProfessionalSeaweedDevourer:
## Two edge cases in the formatting-preservation guarantee

Really like the `ruamel` round-trip approach here — it cleanly fixes the normalization churn. While reviewing I reproduced two cases where the "preserve formatting" guarantee still leaks, and verified a fix for both locally (RED→GREEN, full suite green). Sharing in case they're useful; entirely your call on whether to fold them in.

### 1. CRLF files end up with mixed line endings

`ruamel` always emits `\n`. For a uniformly-CRLF vault file, `dumps()` writes the frontmatter block + delimiters with `\n` while the body keeps `\r\n` — so every merge write produces a mixed-ending file, which is itself a formatting change.

Repro:
```python
content = "---\r\ntitle: hello\r\n---\r\nbody\r\n"
meta, body = frontmatter_io.loads(content)
frontmatter_io.dumps(meta, body)  # -> "---\ntitle: hello\n---\nbody\r\n"  (mixed)
```

### 2. Scalars longer than `width` (4096) get re-wrapped

A long URL/description in frontmatter gets folded on dump, breaking byte-identical preservation:
```python
big = "x" * 5000
meta, body = frontmatter_io.loads(f"---\nurl: {big}\n---\nbody\n")
frontmatter_io.dumps(meta, body) != f"---\nurl: {big}\n---\nbody\n"  # True — value was wrapped
```

### Verified fix

Disable wrapping (and `import sys`):
```python
_YAML.preserve_quotes = True
# Disable line wrapping entirely: any finite width re-folds long scalars
# (URLs, descriptions) on dump, which is exactly the churn this module avoids.
_YAML.width = sys.maxsize
_YAML.indent(mapping=2, sequence=4, offset=2)
```

And in `dumps()`, match the body's line endings so output never mixes:
```python
buf = io.StringIO()
_YAML.dump(metadata, buf)
yaml_text = buf.getvalue()
newline = "\r\n" if "\r\n" in body else "\n"
if newline != "\n":
    yaml_text = yaml_text.replace("\n", newline)
return f"---{newline}{yaml_text}---{newline}{body}"
```

With both applied: CRLF files round-trip byte-identically (no mixed endings), and long scalars stay unwrapped — confirmed end-to-end through `vault_write(merge_frontmatter=True)`. Happy to open a small follow-up PR against your branch if that's easier than applying inline.

I also have a few smaller observations (thread-safety of the shared `_YAML` instance, behavior on malformed existing frontmatter) — will share those separately so this stays focused.

ProfessionalSeaweedDevourer:
## Follow-up: thread-safety + malformed-frontmatter handling

As mentioned above, three more observations — all reproduced and fixed locally (TDD, full suite green). I've put them in a draft branch so you can pull/cherry-pick rather than retype: **ProfessionalSeaweedDevourer/obsidian-web-mcp#2** (stacked on this PR's head, so its diff is just the net-new changes).

### 3. The shared `_YAML` instance isn't thread-safe (most serious)

`_YAML` is a module-level singleton, but `ruamel`'s `YAML` instances hold mutable parser/emitter state and aren't reentrant. The tools are sync functions, and FastMCP runs sync tools in a threadpool, so concurrent `vault_write(merge_frontmatter=True)` calls share one parser. A 16-thread x 400-iteration round-trip stress test produced **6283 errors** (`expected a single document`, duplicate-key, `'NoneType' object has no attribute 'anchor'`). In the merge path these surface as either dropped frontmatter (swallowed `YAMLError`) or a crash.

Fix: build a fresh handler per call.
```python
def _make_yaml() -> YAML:
    yaml = YAML(typ="rt")
    yaml.preserve_quotes = True
    yaml.width = sys.maxsize
    yaml.indent(mapping=2, sequence=4, offset=2)
    return yaml
```
Same stress test after the change: **0 errors, 0 mismatches**. (Construction cost is negligible for human-triggered writes; a thread-local handler is an equivalent alternative.)

### 4 & 5. Malformed frontmatter is handled silently

`loads()` catches `YAMLError` and returns `({}, content)`, conflating "no frontmatter" with "unparseable frontmatter". On merge that means:
- **Malformed *existing* frontmatter** -> existing keys silently dropped, file overwritten with only the new keys.
- **Malformed *new* frontmatter** -> demoted into the body, producing a nested `---` block.

Suggested contract: `loads()` raises on delimiters-present-but-invalid YAML (still returns `({}, content)` when there are no delimiters), and `vault_write` aborts the merge, leaving the file untouched and returning an error, rather than writing something lossy:
```python
except YAMLError as e:
    return json.dumps({"error": f"Frontmatter merge aborted: malformed YAML frontmatter ({e})",
                       "path": path, "written": False})
```
Rationale: for an agent-facing tool, a clear error it can correct beats silently dropping data or emitting a malformed file. This is a deliberate behavior change, so flagging it explicitly — happy to soften to best-effort if you'd prefer.

All of #1–#5 are in the linked draft PR if a single follow-up is easier than applying inline.

jjsmackay:
All five reproduce and are fixed in `016555d` — reimplemented rather than cherry-picked, but the diagnosis on each was spot on, credited you in the commit.

**#3** was the worst, as you flagged: a 16-thread × 300-iter stress test went from thousands of errors (incl. uncaught `AttributeError` escaping `loads()`, not just swallowed `YAMLError`) to zero, via a fresh `_make_yaml()` per call. **#1** — `dumps()` now matches the body's line endings. **#2** — `width = sys.maxsize`. **#4/#5** — went with your strict contract: `loads()` raises on delimiters-present-but-invalid; `vault_write` aborts the merge (file untouched, `written: false`) rather than dropping keys or nesting a stray `---`. Well-formed single-threaded writes are unchanged.

7 new tests cover all of it. Happy to see the other observations you mentioned holding back.

ProfessionalSeaweedDevourer:
Thanks for the thorough reimplementation (and the credit) — your version reads cleaner than my draft, especially re-exporting `YAMLError` from `frontmatter_io` and keeping the merge block flat. Closing my draft branch as superseded.

For the "observations holding back": they're not bugs, just **four merge tests** that aren't covered yet. The behavior is already correct in `016555d`; these just pin it so a future refactor can't regress it silently. Drop-in ready against your current tests:

`tests/test_frontmatter_preservation.py` (uses the existing `vault_dir` / `config` / `vault_write` / `json` already imported there):
```python
def test_merge_overrides_existing_key_value(vault_dir):
    """A new value for an existing key wins, while other keys keep their formatting."""
    path = "fmt.md"
    (config.VAULT_PATH / path).write_text(
        "---\nstatus: 'active'\ntags:\n  - alpha\n  - beta\n---\nbody\n"
    )

    vault_write(path, "---\nstatus: archived\n---\nbody\n", merge_frontmatter=True)

    result = (config.VAULT_PATH / path).read_text()
    assert "status: 'archived'" in result   # value overridden, quote slot kept
    assert "status: 'active'" not in result  # old value gone
    assert "  - alpha" in result             # untouched key keeps block style


def test_merge_bodyless_new_content_keeps_frontmatter(vault_dir):
    """New content with no frontmatter preserves existing frontmatter, replaces body."""
    path = "fmt.md"
    (config.VAULT_PATH / path).write_text(
        "---\ntitle: 'kept'\npinned: yes\n---\nold body\n"
    )

    vault_write(path, "brand new body only\n", merge_frontmatter=True)

    result = (config.VAULT_PATH / path).read_text()
    assert "title: 'kept'" in result
    assert "pinned: yes" in result
    assert "brand new body only" in result
    assert "old body" not in result


def test_no_merge_writes_content_byte_identical(vault_dir):
    """With merge_frontmatter=False the content is written verbatim (no YAML rewrite)."""
    path = "fmt.md"
    (config.VAULT_PATH / path).write_text("---\nold: 1\n---\nold body\n")

    content = "---\nstatus: yes\ntags: [a, b]\nq: 'x'\n---\nbody\n"
    vault_write(path, content, merge_frontmatter=False)

    assert (config.VAULT_PATH / path).read_text() == content
```

`tests/test_frontmatter_io.py` — your `test_update_field_preserves_other_formatting` covers an unquoted value; this adds the quoted case (the interesting one, since the slot's quote style must survive a value swap):
```python
def test_update_existing_quoted_value_keeps_quote_style():
    """Overwriting an existing key's value retains that key's original quote style."""
    metadata, body = frontmatter_io.loads("---\nstatus: 'active'\npriority: 1\n---\nx\n")
    metadata["status"] = "draft"
    out = frontmatter_io.dumps(metadata, body)
    assert "status: 'draft'" in out   # value changed, single-quote slot kept
    assert "priority: 1" in out
```

Why these three integration cases specifically: the override semantic and the bodyless-merge path are the two most common real uses, and `merge_frontmatter=False` is the scope boundary the PR description draws — worth a guard each. That's everything I had; nice work on this.

jimprosser:
The direction here is right and better than what's on `main`: round-tripping with ruamel to preserve comments/quotes/key-order, failing closed on malformed YAML, a fresh handler per call for thread-safety, and the atomic write left intact. One blocker, because it's a data-integrity bug in the exact path this PR is meant to protect.

**Blocking — the frontmatter delimiter regex isn't line-anchored (`frontmatter_io.py:19`).** The closing delimiter is `(?:\r?\n)?---`, so the newline before `---` is optional and an inline `---` inside a value closes the block early. Reproduced:

```
---
title: value --- still title
next: keep
---
body
```

parses as YAML `title: value ` and body `still title\nnext: keep\n---\nbody` — `next: keep` is silently dropped out of the frontmatter into the note body, with no `YAMLError` (so the fail-closed path never triggers). Anchor the closing `---` to its own line (require the preceding newline, or `^---\s*$` with `re.MULTILINE`) and add a test for a value containing `---`.

**Nits.** The abort path returns `{"written": false}` but the success path has no `written` key — align them. The quoted-value-*swap* round-trip (does an existing key's quote slot survive a new value?) is asserted in discussion but not pinned by a committed test. CRLF-on-bodyless-merge falls back to LF.

**Pre-existing, not this PR:** concurrent merges to the same file are still read-modify-write with last-write-wins — worth tracking separately.

Fix the delimiter anchoring and I'll merge.

jimprosser:
I'll take `ruamel.yaml` as a base dependency — preserving frontmatter formatting is close enough to core for a notes server to earn it, and it's light. No need to make it optional.

Two round-trip fixes before it can land (round-trip *is* the point of the PR, so these matter):

1. The closing `---` match isn't line-anchored, so a scalar value containing `---` can be misread as the end of the frontmatter and corrupt the parse.
2. Delimiter line-endings are inferred from the body only, so CRLF frontmatter with an LF/empty body won't round-trip byte-for-byte.

And the tests should hit the hard cases this PR exists to protect — anchors/aliases, key order through `vault_write`, standalone and inter-key comments, a `---`-inside-a-scalar — not just quotes/lists/booleans. Plus a rebase on main. Once those are in, I'll merge.

jimprosser:
Thanks for this — the round-trip preservation is the right idea, and the hardening you've layered on (a per-call `YAML()` for the threadpool, no line-wrap refolding, malformed YAML aborting the merge instead of writing lossily) is exactly the careful kind of work I want. Two things need to change before it can merge, and the first is a data-loss bug.

**1. The frontmatter split silently corrupts or drops YAML (blocker).** The closing-delimiter regex in `frontmatter_io.py` doesn't require `---` to be a whole line, and it doesn't handle a BOM. Verified against this branch:

| input | result |
|---|---|
| `summary: alpha --- beta` | inline `---` read as the closing fence → `{'summary': 'alpha'}`, `status: active` dropped, value mangled |
| a `\|` literal block containing an indented `---` line | truncated at the `---`, later keys dropped |
| UTF-8 BOM before `---` | frontmatter seen as absent → on a merge the existing frontmatter is dropped entirely |

That's worse than the normalization this PR fixes — it's silent data loss on the merge path, and it's a regression against `python-frontmatter`, which handles all three because it only treats `---` as a fence on its own line. Fix: anchor the closing delimiter to a complete line (`^---[ \t]*$` with `re.MULTILINE`, or a line-based split), strip a leading BOM before matching, and decide the no-closing-delimiter case deliberately — I'd fail closed, consistent with the malformed-YAML handling you already added. These all want tests; the current suite only uses `---`-on-its-own-line fixtures, which is why they slipped through.

**2. Rebase on main — this branch predates #38 (blocker).** `write.py` here uses `json.dumps` throughout, while current main uses `serialization.dumps` (the non-ASCII UTF-8 fix from #38, already merged). Merging as-is reverts that — non-ASCII tool responses go back to `\uXXXX` escaping. Please rebase and route the new code, including the malformed-YAML error return, through `serialization.dumps`. (While there: that error response uses `"written": False`; the rest of the tool uses `created` — worth aligning.)

Smaller, non-blocking:
- CRLF newline is inferred from the body, not the original frontmatter, so a CRLF file with an empty body (or CRLF existing + LF new body) can come back with LF frontmatter. A "preserve formatting" change should probably nail this.
- Carrying ruamel (write) alongside python-frontmatter (read/index) means a `yes`/`no` value reads back as a boolean on the read path but stays textual on the write path — clients can see different types depending on which path they hit. Not a merge bug, but worth a note.

Dependency wiring is correct — `ruamel.yaml>=0.18` as a base runtime dep is the call I wanted. Fix the split bug and the rebase and I think this is close.

jjsmackay:
Both blockers are fixed and it's rebased clean on `main`.

**1. Split bug — line-anchored delimiter + BOM (`9eb5ecc`).** Replaced the delimiter regex with a line-based split: only a non-indented `---` line (trailing spaces/tabs and CR tolerated) opens or closes the block, so neither a `---` inside a scalar value nor an indented `---` in a literal block can be misread as the closing fence. A leading UTF-8 BOM is stripped before matching, so a BOM-prefixed file is no longer seen as having no frontmatter — which previously dropped it entirely on a merge. An opening `---` with no closing fence now fails closed (`YAMLError` → merge aborts), consistent with the malformed-YAML handling.

**2. Rebase + serialization (`a9ae857`).** Rebased onto current `main`; the new merge code and the malformed-YAML abort now route through `serialization.dumps`, so #38's non-ASCII handling is preserved. Aligned the abort response from `"written": false` to `"created": false` to match the rest of the tool.

**CRLF inferred from body (non-blocking, done).** `loads` now records the source file's line ending and `dumps` honours it, so an all-frontmatter file with an empty/LF body round-trips byte-identically instead of falling back to LF.

**Test coverage.** Pinned the hard cases you listed, not just quotes/lists/booleans: `---` inside a scalar and an indented `---` in a literal block (both also end-to-end through `vault_write`), BOM prefix, unterminated frontmatter, anchors/aliases, standalone and inter-key comments, key order preserved through a merge, CRLF-with-empty-body round-trip, and a quoted-value swap. Also added the override / bodyless / no-merge integration cases from the earlier review.

**ruamel/python-frontmatter type divergence (non-blocking).** Left as-is for now: `yes`/`no` reading back as a bool on the python-frontmatter index path while staying textual on the ruamel write path is a real inconsistency, but reconciling it means putting both paths on one loader — bigger than this PR. Happy to track it separately.

I've left the pre-existing concurrent same-file last-write-wins merge untouched, as you flagged — worth its own issue.

Issue #19, opened 2026-05-11
MCP endpoint mounted at /mcp by default — Claude.ai connector fails with 404


matt2053:
FastMCP's \`streamable_http_path\` defaults to \`/mcp\`, so the MCP endpoint is reachable at \`https://your-host/mcp\`. Claude.ai's connector probes \`POST /\` and, after completing OAuth successfully, hits \`POST /\` with the bearer token — getting 404 and reporting "Authorization failed."

Context: discovered while setting up the server as a Claude.ai custom connector (Settings → Connectors → Add custom connector).

**Fix:** pass \`streamable_http_path="/"\` to \`FastMCP()\` in \`server.py\`:

\`\`\`python
mcp = FastMCP(
    "obsidian_web_mcp",
    stateless_http=True,
    json_response=True,
    streamable_http_path="/",   # ← add this
    ...
)
\`\`\`

Tested with Claude.ai web connector and Claude iOS app.

PR #43, opened 2026-06-09
Make MCP mount path configurable (default /)

### LINKED ISSUE #46: Allow configuring the MCP mount path (hosting behind a path prefix)
`streamable_http_path` is hardcoded to `/` (for connector interop, #19), so the server can't be hosted under a path prefix like `https://host/mcp` alongside other services on the same hostname.

Cloudflare Tunnel can't work around this: cloudflared ingress matches paths but cannot rewrite or strip them (cloudflare/cloudflared#563), and rejects a path in the `service:` URL. So the prefix must be handled in the application or a separate reverse proxy.

Request: read the path from an env var defaulting to `/`, leaving the connector-interop default unchanged while enabling path-prefix deployments. Happy to PR (single concern, tests, default preserved). Two wrinkles the PR handles, both gated on `VAULT_MCP_PATH != "/"` so the default stays byte-identical: (1) the MCP-spec `GET/HEAD /` probe is only mounted off-root, or it shadows the transport; (2) that probe must answer without credentials, so bearer auth exempts `GET/HEAD /` under the same guard — at the default `/` the exempt set is empty and root stays fully authenticated. Flagging the auth touch up front since it's security-adjacent; happy to split it into a follow-up if you'd rather keep this PR to the path config alone.

jjsmackay:
Closes #46.

**Problem:** `streamable_http_path` is hardcoded to `/` (#19), so the server can't sit under a prefix like `/mcp`. cloudflared can't strip paths (cloudflare/cloudflared#563), so the prefix has to come from the app.

**Fix:** `VAULT_MCP_PATH`, default `/`. Default stays byte-identical; #19's root mount holds.

Both gated on `VAULT_MCP_PATH != "/"`:
- Mount the MCP-spec `GET/HEAD /` probe so root still answers once the transport moves off it.
- `auth.py` exempts `GET/HEAD /` from bearer auth. Empty set at `/`, so root stays authenticated by default.

**Tests:** 76 pass, 4 new (`test_mcp_path.py`). The probe route mounts in `main()`; verified by hand.

The `auth.py` exemption is the security-adjacent touch flagged in #46 — happy to split it into a follow-up if you'd rather keep this PR to the path config alone.

### DIFF
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -19,6 +19,14 @@ _AUTH_EXEMPT_PATHS = {
     "/oauth/register",
 }
 
+# (method, path) pairs exempt from auth. The MCP spec 2025-06-18 probe on / must
+# answer GET/HEAD without credentials. This is ONLY active when MCP is mounted off
+# root (VAULT_MCP_PATH != "/"); when MCP is at root the transport owns GET/HEAD /
+# and must stay fully authenticated, so the set is empty and behaviour is unchanged.
+_AUTH_EXEMPT_METHOD_PATHS = (
+    {("GET", "/"), ("HEAD", "/")} if config.VAULT_MCP_PATH != "/" else set()
+)
+
 
 def _www_authenticate(request: Request, error: str) -> str:
     """RFC 9728 challenge header pointing clients at the protected-resource metadata.
@@ -40,6 +48,9 @@ class BearerAuthMiddleware(BaseHTTPMiddleware):
 
     async def dispatch(self, request: Request, call_next):
         if request.url.path in _AUTH_EXEMPT_PATHS:
+            return await call_next(request)
+
+        if (request.method, request.url.path) in _AUTH_EXEMPT_METHOD_PATHS:
             return await call_next(request)
 
         if not VAULT_MCP_TOKEN:
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -6,6 +6,13 @@ VAULT_PATH = Path(os.environ.get("VAULT_PATH", os.path.expanduser("~/Obsidian/My
 VAULT_MCP_TOKEN = os.environ.get("VAULT_MCP_TOKEN", "")
 VAULT_MCP_PORT = int(os.environ.get("VAULT_MCP_PORT", "8420"))
 
+# HTTP path the MCP transport is mounted at. Defaults to "/" so connectors that
+# POST to the root complete the handshake (#19) -- changing this default would
+# break that, so leave it unless you deliberately host under a path prefix.
+# Setting it (e.g. "/mcp") lets the server live alongside other services on one
+# hostname behind a reverse proxy that cannot rewrite paths (Cloudflare Tunnel).
+VAULT_MCP_PATH = os.environ.get("VAULT_MCP_PATH", "/")
+
 # OAuth 2.0 client credentials (for Claude app integration)
 VAULT_OAUTH_CLIENT_ID = os.environ.get("VAULT_OAUTH_CLIENT_ID", "vault-mcp-client")
 VAULT_OAUTH_CLIENT_SECRET = os.environ.get("VAULT_OAUTH_CLIENT_SECRET", "")
--- a/src/obsidian_vault_mcp/server.py
+++ b/src/obsidian_vault_mcp/server.py
@@ -17,6 +17,7 @@ from .config import (
     VAULT_MCP_ALLOWED_HOSTS,
     VAULT_MCP_FORWARDED_ALLOW_IPS,
     VAULT_MCP_HOST,
+    VAULT_MCP_PATH,
     VAULT_MCP_PORT,
     VAULT_MCP_TOKEN,
     VAULT_PATH,
@@ -46,9 +47,10 @@ mcp = FastMCP(
     "obsidian_web_mcp",
     stateless_http=True,
     json_response=True,
-    # Serve MCP at "/" (not the /mcp default) so connectors that POST to the root
-    # complete the handshake instead of 404ing (#19).
-    streamable_http_path="/",
+    # Mount path for the MCP transport. Defaults to "/" (via VAULT_MCP_PATH) so
+    # connectors that POST to the root complete the handshake instead of 404ing
+    # (#19); set VAULT_MCP_PATH to host under a prefix like "/mcp".
+    streamable_http_path=VAULT_MCP_PATH,
     lifespan=lifespan,
     transport_security=TransportSecuritySettings(
         enable_dns_rebinding_protection=True,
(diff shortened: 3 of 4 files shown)


ProfessionalSeaweedDevourer:
Reviewed this with a few focused agents and then ran some experiments against the branch to sanity-check the parts that looked risky. Overall the design seems solid: fail-closed `sys.exit(1)` is preserved, the default (`VAULT_MCP_PATH="/"`) stays byte-identical (empty exemption set, no probe route, root fully authenticated), the exemption is method-scoped so `POST /` still authenticates, and I didn't see any route shadowing. Nice work.

On the unvalidated `VAULT_MCP_PATH`: I went in expecting a possible auth bypass, but from the cases I tried on my end it held up okay, so I'd lean toward this not being a blocker:

- `""`, `" "`, `"mcp"` (empty / whitespace / no leading slash): `mcp.streamable_http_app()` raises `AssertionError: Routed paths must start with '/'`, which `main()`'s `except Exception` catches and turns into `sys.exit(1)`. So in my testing these failed closed at startup before the exemption set could matter.
- `"/mcp/"` (trailing slash) was the only bad value I found that gets silently accepted, and it looked like an availability issue rather than a security one: the data plane is `POST` to the transport (mounted at `/mcp/`) and stays authenticated, while the only exempt route is `GET/HEAD /`, which serves the static protocol-version probe and no vault data. The visible symptom is just a connector POSTing to `/mcp` getting a 404, i.e. the #19 handshake failure resurfacing.

Caveat that this is only as good as the cases I happened to try, so take it as "nothing jumped out" rather than a guarantee. Two small, non-blocking suggestions:

## A short note on the accepted path format

The trailing-slash case starts a silently-broken server (connector 404s with no startup error), so a one-line note in the `config.py` comment or README might save the next operator some debugging: `VAULT_MCP_PATH` should be `/` or a leading-slash path with no trailing slash (e.g. `/mcp`).

Possibly worth pairing with an operational caveat: changing this on a deployment that already has connected clients seems to be effectively a breaking change. A connector that cached the old path has to be deleted and re-added, since neither side renegotiates the mount point. (Ran into this myself right after the v0.2.0 release: Claude web was still holding `/mcp` while the server served `/`, and only delete-and-reconnect sorted it out.)

## Behavioral tests for the probe + exemption look cheap to add

The note that the probe route "needs a running app, so it is not unit-tested" may be avoidable: `tests/test_auth.py` already drives `BearerAuthMiddleware` with Starlette's `TestClient`, no server needed. Lifting `mcp_root_probe` to a module-level function makes it directly testable, and the security-relevant contract might be worth locking in behaviorally:

- default mode: unauthenticated `GET /` -> 401
- off-root mode: `GET/HEAD /` -> 200 with `MCP-Protocol-Version`, but `POST /` -> 401 and `GET /other` -> 401 (so the bypass stays keyed on `(method, path)`, not path)

I have a version of these working locally (10 small tests, full suite still green); happy to send the diff or a PR against your branch if that's useful.

## Minor: the "MCP spec 2025-06-18 probe" comment

As far as I can tell the 2025-06-18 Streamable HTTP spec doesn't define a "GET / probe that answers with the protocol version" — `MCP-Protocol-Version` is a header the *client* sends, and a GET on the transport endpoint normally opens an SSE stream or returns 405. The route still seems like a useful connector affordance, just maybe not spec-mandated, so it might read more accurately framed that way rather than as "spec ... probe".

ProfessionalSeaweedDevourer:
Follow-up to the above, purely optional and separable from the docs/tests suggestions.

If you'd rather enforce the path format in code than rely on a README note, here's the small version I tried locally. The framing is fail-fast UX and pinning the contract, not fixing a security hole (per the experiments above I don't think there is one):

```python
VAULT_MCP_PATH = os.environ.get("VAULT_MCP_PATH", "/").strip()
if VAULT_MCP_PATH != "/" and (
    not VAULT_MCP_PATH.startswith("/") or VAULT_MCP_PATH.endswith("/")
):
    raise ValueError(
        "VAULT_MCP_PATH must be '/' or a '/'-prefixed path without a trailing "
        f"slash (e.g. '/mcp'); got {os.environ.get('VAULT_MCP_PATH')!r}"
    )
```

Two reasons it might be worth the four lines:

1. **Turns the one silently-bad case into a loud one.** `"/mcp/"` currently boots a server whose connector just 404s; this makes it exit at startup with a message that names the cause.
2. **Pins behavior that's currently emergent.** Today the empty/no-slash cases fail closed only because Starlette happens to assert `Routed paths must start with '/'`. If FastMCP ever normalized those (e.g. coerced `""` to `/`), that safety would change silently. Validating here keeps the transport mount and the auth exemption reading the same sanitized value regardless of upstream behavior.

Reasons it might not be worth it: it moves `config.py` from pure constants to "can raise at import," which couples every test that imports config to a valid env, and the failure modes are operator-controlled and self-inflicted, the same class as a wrong port or vault path that the project doesn't otherwise validate.

One limitation to be honest about: this only checks leading/trailing slashes. It would not catch mounting the transport on an already-exempt path (e.g. `VAULT_MCP_PATH=/health` or `/oauth/token`), which is a different and more exotic footgun. I don't think it's worth guarding here, but flagging it so the validation isn't mistaken for total.

Happy to open this as a tiny PR against your branch if you want it, or leave it as just the README note from the previous comment, your call.

jimprosser:
Thanks for this — defaulting to `/` keeps it byte-identical to today, it's small, and the probe is data-free. One blocker before it can merge, plus a nit.

**Blocking — auth bypass via an unvalidated mount path.** `VAULT_MCP_PATH` isn't validated, and the bearer-auth exemptions match by path only. If the value lands on an already-exempt path, the MCP transport mounts on an unauthenticated route. Verified against a live server: with `VAULT_MCP_PATH=/health`, an unauthenticated `POST /health` reaches MCP `initialize`/tool calls, while `POST /mcp` correctly returns 401. Same with `/.well-known/oauth-authorization-server`. A bad value fails *open*, which is the wrong direction for a setting whose whole purpose is to be changed.

Fix: validate `VAULT_MCP_PATH` at startup against the set of auth-exempt paths (reject any collision), and reject a non-leading-slash or trailing-slash value — fail closed with a clear error rather than booting.

**Nit.** A trailing slash (`/mcp/`) currently boots a silently-broken server (connector 404s) — caught by the same validation above. Also the code comment calling the `GET /` probe "spec-mandated" overstates the 2025-06-18 Streamable HTTP spec; `MCP-Protocol-Version` is a client-sent header. Cosmetic.

Once the path is validated against the exempt set, I'm happy to merge.

jimprosser:
Merged to `main` in dd6af28 (your authorship preserved). I added a maintainer hardening commit on top: `VAULT_MCP_PATH` is now validated at startup and fails closed if it's malformed or collides with an auth-exempt route (`/oauth/*`, `/health`, `/.well-known/*`), so it can't end up serving the vault unauthenticated; the app assembly is extracted into `build_app()` with end-to-end TestClient tests for the auth seam (default rejects unauth; off-root `/` is liveness-only; transport stays behind bearer auth); README documents the env var. Ran it through two Codex security passes — no auth bypass found.

Closing since the squashed commit differs from this branch's head. Clean, well-scoped contribution — thanks.

PR #44, opened 2026-06-09
Optional post-write hook for external sync/backup/webhooks

### LINKED ISSUE #47: Optional post-write hook for external sync/backup/webhooks
There's no way to react to vault mutations (e.g. git-commit the vault, trigger a backup, notify a webhook) without polling. Proposal: an optional, generic post-write hook — if `VAULT_MCP_POST_WRITE_CMD` is set, run it fire-and-forget after each mutation, injecting `MCP_OPERATION` and `MCP_PATHS` into the environment. No-op when unset; off by default. Happy to PR with tests. Flagging up front: it executes an operator-configured command, so the docs would note that clearly.

jjsmackay:
Closes #47.

**Problem:** nothing reacts to a mutation without polling (git sync, backup, webhook).

**Fix:** optional dispatcher. Set `VAULT_MCP_POST_WRITE_CMD` and it runs after each successful mutation, fire-and-forget in a daemon thread, with:
- `MCP_OPERATION`: created / updated / moved / deleted
- `MCP_PATHS`: colon-separated vault-relative paths

No-op when unset (default). Daemon thread with a timeout, logs on failure, so it can't block or fail a tool call. Generic runner, not a single-use flag.

**Call sites:** success path of `vault_write` / `vault_edit` / `vault_append` / `vault_batch_frontmatter_update` (write.py), `vault_move` / `vault_delete` (manage.py).

**Security:** runs an operator-configured shell command. Left the README alone; will add a docs note where you prefer.

**Tests:** 76 pass, 4 new (`test_hooks.py`): env injection, no-op unset, no-op empty paths, thread spawn.

### DIFF
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -74,6 +74,13 @@ def advertised_base_url(request_base_url: str) -> str:
     """
     return (VAULT_MCP_PUBLIC_URL or request_base_url).rstrip("/")
 
+# Optional post-write hook. When set, this shell command runs fire-and-forget in
+# a daemon thread after each successful vault mutation, with MCP_OPERATION (e.g.
+# "created"/"updated"/"moved"/"deleted") and MCP_PATHS (colon-separated, vault-
+# relative) injected into its environment. Empty = disabled (no-op, the default).
+# It executes an operator-configured command, so only set it to something you trust.
+VAULT_MCP_POST_WRITE_CMD = os.environ.get("VAULT_MCP_POST_WRITE_CMD", "")
+
 # Safety limits
 MAX_CONTENT_SIZE = 1_000_000  # 1MB max write size
 MAX_BATCH_SIZE = 20           # Max files per batch operation
--- a/src/obsidian_vault_mcp/hooks.py
+++ b/src/obsidian_vault_mcp/hooks.py
@@ -1,0 +1,83 @@
+"""Generic post-write hook dispatcher.
+
+After any vault mutation, tools call fire_post_write() with a human-readable
+operation string and the list of affected vault-relative paths.
+
+If ``VAULT_MCP_POST_WRITE_CMD`` is set, that shell command is executed
+fire-and-forget in a daemon thread.  Two environment variables are injected:
+
+    MCP_OPERATION   — e.g. "created", "updated", "deleted"
+    MCP_PATHS       — colon-separated vault-relative paths
+
+If the variable is not set, fire_post_write() is a no-op.
+"""
+
+import logging
+import os
+import subprocess
+import threading
+from collections.abc import Sequence
+
+from .config import VAULT_MCP_POST_WRITE_CMD
+
+logger = logging.getLogger(__name__)
+
+_PATH_SEP = ":"
+
+
+def _run_cmd(cmd: str, operation: str, paths: list[str]) -> None:
+    """Execute the configured post-write shell command.
+
+    Called inside a daemon thread — never raises, only logs.
+    """
+    env = os.environ.copy()
+    env["MCP_OPERATION"] = operation
+    env["MCP_PATHS"] = _PATH_SEP.join(paths)
+
+    try:
+        result = subprocess.run(
+            cmd,
+            shell=True,
+            env=env,
+            capture_output=True,
+            text=True,
+            timeout=120,
+        )
+        if result.returncode != 0:
+            logger.warning(
+                "post-write hook exited %d: %s",
+                result.returncode,
+                result.stderr.strip(),
+            )
+        else:
+            logger.debug("post-write hook ok: %s %s", operation, paths)
+    except subprocess.TimeoutExpired:
+        logger.warning("post-write hook timed out: %s %s", operation, paths)
+    except Exception as exc:
+        logger.warning("post-write hook error: %s", exc)
+
+
+def fire_post_write(operation: str, paths: Sequence[str]) -> None:
+    """Dispatch a post-write hook fire-and-forget.
+
+    Does nothing if ``VAULT_MCP_POST_WRITE_CMD`` is not configured.
+
+    Args:
+        operation: Human-readable verb describing the mutation
+                   (e.g. "created", "updated", "deleted", "moved").
+        paths:     Vault-relative paths of the affected files.
+    """
+    if not VAULT_MCP_POST_WRITE_CMD:
+        return
+
+    path_list = list(paths)
+    if not path_list:
+        return
+
+    t = threading.Thread(
+        target=_run_cmd,
+        args=(VAULT_MCP_POST_WRITE_CMD, operation, path_list),
+        daemon=True,
+        name="mcp-post-write",
+    )
+    t.start()
--- a/src/obsidian_vault_mcp/tools/manage.py
+++ b/src/obsidian_vault_mcp/tools/manage.py
@@ -3,6 +3,7 @@
 import json
 import logging
 
+from ..hooks import fire_post_write
 from ..vault import list_directory, move_path, delete_path, resolve_vault_path
 
 logger = logging.getLogger(__name__)
@@ -38,6 +39,7 @@ def vault_move(source: str, destination: str, create_dirs: bool = True) -> str:
     """Move a file or directory within the vault."""
     try:
         moved = move_path(source, destination, create_dirs=create_dirs)
+        fire_post_write("moved", [source, destination])
         return json.dumps({"source": source, "destination": destination, "moved": moved})
     except ValueError as e:
         return json.dumps({"error": str(e), "source": source, "destination": destination})
@@ -56,6 +58,7 @@ def vault_delete(path: str, confirm: bool = False) -> str:
 
     try:
         deleted = delete_path(path)
+        fire_post_write("deleted", [path])
         return json.dumps({"path": path, "deleted": deleted})
     except ValueError as e:
         return json.dumps({"error": str(e), "path": path})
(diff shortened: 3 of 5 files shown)


ProfessionalSeaweedDevourer:
Thanks for this contribution. The call-site discipline here is genuinely solid, and the disabled-by-default design with the trust caveat in `config.py` is exactly right. I ran a multi-pass review (general correctness, error handling, test coverage, doc accuracy) and then **empirically reproduced every finding at the PR head (a4c90d8) in an isolated worktree** (macOS, CPython 3.12, real vault directory, real hook shell commands; no mocks except where explicitly noted). Repro details inline.

> **Review process note (correction on record):** an earlier draft of this review claimed that after a hook timeout, `subprocess.run`'s internal `communicate()` could block forever on pipes held by orphaned grandchildren, suppressing even the timeout warning. Empirical testing **refuted** that claim on CPython 3.12 POSIX: `TimeoutExpired` was raised at exactly 3.0s with a replicated call (timeout constant shortened 120 to 3 for the experiment), and the warning path works. CPython's POSIX timeout path calls `wait()`, not `communicate()`, after the kill (that branch is Windows-only). The claim has been withdrawn; the orphaned-process finding below stands on its own evidence.

## Blocking

### 1. `fire_post_write` itself can raise into the tool-call path

The "never raises" guard lives in `_run_cmd` (inside the thread), but `Thread(...).start()` in `fire_post_write` can raise `RuntimeError: can't start new thread` under resource pressure, and nothing catches it. I verified the propagation by injecting that exact exception into `Thread.start` (fault injection at the one OS boundary; everything downstream is real code on a real vault):

- `vault_write` wrote the file to disk, then returned `{"error": "can't start new thread"}`. The mutation succeeded; the response says it failed.
- An agent retrying `vault_append` after that phantom failure **appended the content a second time**: final file content was `'hello\n\nTODO item\n\nTODO item'`.
- `vault_edit` changed the file on disk and returned `"changed": false`.
- `vault_batch_frontmatter_update` is worse: the call at write.py:255 sits outside any try/except, so the exception escaped the tool function entirely and all per-file results were lost (the file on disk was updated).

Fix is one guard: wrap the body of `fire_post_write` (at minimum `t.start()`) in `try/except Exception` with a `logger.warning(..., exc_info=True)`. That moves the module's stated contract ("must never block or fail a tool call") to the function that actually needs it.

### 2. `MCP_PATHS` colon separator collides with legal filenames

`resolve_vault_path` rejects null bytes, dot components, and vault escapes, but not colons, and a colon is a legal filename character on macOS and Linux. Reproduced end to end with a real hook command recording its environment:

```
vault_write("notes/standup 10:30.md")  ->  MCP_PATHS='notes/standup 10:30.md'
    consumer split(':')  ->  ['notes/standup 10', '30.md']   (1 real path, 2 parsed)
vault_move("notes/a.md", "archive/b:c.md")  ->  MCP_PATHS='notes/a.md:archive/b:c.md'
    consumer split(':')  ->  3 tokens for 2 paths, boundaries unrecoverable
```

The colliding file was created by this server itself via `vault_write`, so this is reachable through the documented API, not a hypothetical. Options, any of which resolves it: NUL separator (`\0` is already guaranteed absent from every path by `resolve_vault_path`, cleanest), newline, or indexed vars (`MCP_PATH_0..N` plus `MCP_PATH_COUNT`). At minimum, document the limitation next to the env var.

## Important

### 3. A misconfigured hook is a permanent, invisible no-op

I started the actual server with `VAULT_MCP_POST_WRITE_CMD` set: the full INFO-level startup log contains zero mention of the hook. A typo'd env var name, a wrong shell, or a process manager dropping the variable all produce a hook that never fires, with log output identical to a working one (success is logged at DEBUG only). For the sync/backup use case this README invites, the operator discovers it at restore time. One startup line ("post-write hook enabled: ...") closes most of this.

### 4. Timeout leaves orphaned grandchildren; no process-group kill

Reproduced with the exact `subprocess.run` call shape from `_run_cmd` (timeout 120 shortened to 3): on `TimeoutExpired`, only the immediate `/bin/sh` is killed; a backgrounded grandchild (`sleep 6360 &`) survived and kept running unkilled and unlogged. A compound hook (`rsync ... && git push`, pipelines) leaks its workers on every timeout. Standard fix: `start_new_session=True` plus `os.killpg` on timeout. Related: there is no concurrency control, so rapid sequential mutations run hooks concurrently, and two concurrent `git commit` hooks will fight over the index; a single worker queue would serialize them. Also, the 120s ceiling is hardcoded, not mentioned in any operator-facing doc, and the timeout warning does not state the limit; consider `VAULT_MCP_POST_WRITE_TIMEOUT`.

### 5. Daemon threads silently drop hooks at interpreter exit

Reproduced with a control: a process that calls `vault_write` and exits immediately leaves the file on disk but the 0.4s hook never ran (its marker file was never created); the identical run lingering 1.5s before exit completed the hook. So a mutation landing just before SIGTERM/restart is never synced, with no log, and an in-flight hook can be killed mid-run (mid `git commit`). The mechanism is disclosed ("daemon thread") but the consequence is not. Minimum: document the shutdown caveat. Better: track threads and `join(timeout=...)` on shutdown.

### 6. The integration layer is not pinned by any test

Mutation test: I deleted all six `fire_post_write(...)` call lines from write.py and manage.py and the full suite still passed 76/76. The unit tests in `test_hooks.py` are good as far as they go, but none of the contracts the diff actually adds are covered: created vs updated selection in `vault_write`/`vault_append`, no fire on `vault_edit` dry_run or no-change, batch collecting only successful paths, `vault_move` passing both paths, `vault_delete` firing only on success. A recorder patched at the tool-module import seam (`monkeypatch.setattr("obsidian_vault_mcp.tools.write.fire_post_write", recorder)`) over the existing vault fixture covers all of these in 7 or 8 short tests, matching the behavior-test style of `test_tools.py`.

## Minor

- `_run_cmd`'s non-zero-exit warning logs `stderr` but not `stdout` (where many tools report errors), and omits operation/paths while the timeout branch includes them; the generic `except` discards the traceback (`exc_info=True` would keep it). Unbounded stderr is also dumped raw into the log; consider truncating.
- Docstrings: "never raises, only logs" is slightly overbroad (env construction sits before the `try`); the empty-paths no-op is undocumented in both the module docstring and `fire_post_write` (and `vault_batch_frontmatter_update` relies on it); the module docstring's operation examples omit `"moved"`.
- Test mocks are coupled to call signatures: `fake_run(cmd, shell, env, capture_output, text, timeout)` and `FakeThread.__init__(self, target, args, daemon, name)` break on any added kwarg with no behavior change; `**kwargs` keeps them honest. The `spawned["args"]` tuple assertion pins `_run_cmd`'s positional signature.
- Pre-existing quirk this PR makes observable (reproduced): `vault_append("ghost.md", "")` on a nonexistent file returns `"created": true` while creating no file and firing no hook. Not caused by this PR, but a sync consumer cross-checking responses against hook events will see a phantom create; worth a note.

## What is solid (verified, not just read)

- Success-path discipline checks out everywhere I probed: `vault_edit` dry_run and no-op edits never fire, `vault_append` reuses the precomputed `created` flag and fires only when content changed, an all-failed batch fires nothing, and `move_path`/`delete_path` always return `True` or raise, so firing after they return is safe.
- Disabled-by-default, the empty-paths guard, and per-call thread isolation with immutable args are all correct.
- The `config.py` comment block is accurate and states the trust caveat plainly.

Happy to discuss any of these; items 1 and 2 are the ones I would consider blocking, and both are small diffs.

jimprosser:
This is genuinely careful work — passing the operation/paths through environment variables instead of building a shell string means it is *not* command-injectable (a path like `notes/$(rm -rf ~).md` lands inert in `$MCP_PATHS`), it's off by default, and the call-site discipline (dry-run/no-op skip, batch fires only on successful paths, move passes both) is right. Three blockers before it can merge on a server that holds a personal vault.

**Blocking 1 — the hook inherits all server secrets.** `env = os.environ.copy()` hands every hook process (and anything it execs) `VAULT_MCP_TOKEN` and the OAuth secrets it has no need for. A "backup/webhook helper" shouldn't receive vault-takeover credentials. Pass a scrubbed env: `MCP_OPERATION` + `MCP_PATHS` plus an explicit passthrough allowlist, not the whole environment.

**Blocking 2 — a successful write can be reported as failed.** The "never fails the tool call" guarantee lives in `_run_cmd` (inside the thread), but `t.start()` in `fire_post_write` is unguarded. A `Thread.start()` RuntimeError propagates *after* the file is already written — so the tool returns an error (and `vault_batch_frontmatter_update` loses all per-file results, since the call site at `write.py:255` is outside the per-file try/except). An agent that retries then double-appends. Wrap `t.start()` and log-and-continue.

**Blocking 3 — no operator docs.** The feature, its env vars, the 120s timeout, and the trust model aren't in the README. For a hook that runs shell commands on every vault write, the trust contract needs to be documented before someone enables it.

Also worth addressing: the 120s timeout kills only the shell wrapper, not the process group, so a compound hook (`rsync && git push`) can orphan children that pile up; `MCP_PATHS`'s colon separator collides with colons in filenames (a `\0` separator avoids it); the wiring is untested (deleting every `fire_post_write` call still passes the suite — add a test at the tool seam); and there's no write rate limit, so an unbounded thread+subprocess per write is a DoS lever.

Not a backdoor and the injection-safety is real — but please address the three blockers first.

jimprosser:
Thanks for this — it's cleanly built and off-by-default, and I went back and forth. I'm going to decline it, and I want to be straight about why. The governing question for this project is "does a feature earn its attack surface on a server exposed through a tunnel that holds a private vault," and a generic "run an operator-configured command after every write" is the one capability where my answer is no — independent of how carefully it's implemented. It's the highest-leverage thing for an attacker to reach if anything upstream ever slips, and the people running this are the same people who'd footgun it. So it's a threat-model call, not a code-quality one.

Not a knock on the work — #43 and #45 are exactly what I want, and I'd rather have your attention there. Closing this and #47. Thanks again.

Issue #28, opened 2026-06-03
Requests timing out - appears to be due to building an index in critical path


tleyden:
I'm getting these connection errors from the ChatGPT desktop app:

> Still timing out — the MCP endpoint itself appears unavailable right now:
> Connection timed out

I'm running it with a large Obsidian Vault on a VPS with fairly low CPU resources, and I notice during the connection period the CPU usage spikes to over 100%, and I see this in the following logs:

```
Starting vault MCP server. Vault: /path/to/obsidian_vaults                                                                                                                                                                                                                      server.py:27
[06/03/26 15:56:07] INFO     Frontmatter index built: 2269 files in 32.06 seconds                                                                                                                                                                                                                      frontmatter_index.py:42
[06/03/26 15:56:10] INFO     Frontmatter index built: 2269 files indexed                                                                                                                                                                                                                                          server.py:29
                    INFO     Terminating session: None                                                                                                                                                                                                                                                  streamable_http.py:785
                    INFO     Vault MCP server shut down.
```

Is synchronously building an index before it responds, causing the timeout?

tleyden:
Workaround: I restarted the process, tried again, and this time it worked.

In the logs, the indexing seemed to go much faster

```
INFO:     Uvicorn running on http://0.0.0.0:3001 (Press CTRL+C to quit)
[06/03/26 16:02:04] INFO     Starting vault MCP server. Vault: /path/to/obsidian_vaults                                                                                                                                                                                                                  server.py:27
                    INFO     Frontmatter index built: 2269 files in 0.14 seconds                                                                                                                                                                                                                       frontmatter_index.py:42
                    INFO     Frontmatter index built: 2269 files indexed                                                                                                                                                                                                                                          server.py:29
INFO:     172.71.166.153:0 - "POST /mcp HTTP/1.1" 200 OK
                    INFO     Terminating session: None                                                                                                                                                                                                                                                  streamable_http.py:785
                    INFO     Vault MCP server shut down.                                                                                                                                                                                                                                                          server.py:32
[06/03/26 16:02:05] INFO     Starting vault MCP server. /path/to/obsidian_vaults                                                                                                                                                                                                                      server.py:27
                    INFO     Frontmatter index built: 2269 files in 0.14 seconds                                                                                                                                                                                                                       frontmatter_index.py:42
                    INFO     Frontmatter index built: 2269 files indexed                                                                                                                                                                                                                                          server.py:29
[...]

PR #45, opened 2026-06-09
Optional heartbeat to a push URL for uptime monitoring

### LINKED ISSUE #48: Optional heartbeat to a push URL for uptime monitoring
There's no liveness signal for push-style uptime monitors (e.g. a Healthchecks/Uptime Kuma push URL). Proposal: an optional heartbeat — if `VAULT_MCP_HEARTBEAT_URL` is set, a daemon thread GETs it every `VAULT_MCP_HEARTBEAT_INTERVAL` seconds (default 60); failures are logged and swallowed so a flaky monitor can't take the server down. No-op when unset; off by default. Started in `main()` for the process lifetime rather than the per-request lifespan, which under `stateless_http` would otherwise spawn one heartbeat per session (same reasoning as the index, #28). Happy to PR with tests.

jjsmackay:
Closes #48.

**Problem:** no liveness signal for push-style uptime monitors.

**Fix:** set `VAULT_MCP_HEARTBEAT_URL` and a daemon thread GETs it every `VAULT_MCP_HEARTBEAT_INTERVAL`s (default 60). No-op when unset. Failed pings get logged and swallowed, so a flaky monitor can't take the server down.

The thread starts in `main()` (process lifetime), not the per-request lifespan; with `stateless_http=True` that would spawn one heartbeat per session (same reasoning as the index, #28).

`_heartbeat_ping` is split from the loop, so the GET and the error-swallowing loop test without a network or a real sleep.

**Tests:** 74 pass, 2 new (`test_heartbeat.py`).

### DIFF
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -74,6 +74,13 @@ def advertised_base_url(request_base_url: str) -> str:
     """
     return (VAULT_MCP_PUBLIC_URL or request_base_url).rstrip("/")
 
+# Optional liveness heartbeat. When VAULT_MCP_HEARTBEAT_URL is set, the server GETs
+# it every VAULT_MCP_HEARTBEAT_INTERVAL seconds from a daemon thread, for push-style
+# uptime monitors (Uptime Kuma, Healthchecks.io, Cronitor, ...). Empty = disabled
+# (the default); failures are logged, never fatal.
+VAULT_MCP_HEARTBEAT_URL = os.environ.get("VAULT_MCP_HEARTBEAT_URL", "")
+VAULT_MCP_HEARTBEAT_INTERVAL = int(os.environ.get("VAULT_MCP_HEARTBEAT_INTERVAL", "60"))
+
 # Safety limits
 MAX_CONTENT_SIZE = 1_000_000  # 1MB max write size
 MAX_BATCH_SIZE = 20           # Max files per batch operation
--- a/src/obsidian_vault_mcp/server.py
+++ b/src/obsidian_vault_mcp/server.py
@@ -8,6 +8,9 @@ import atexit
 import json
 import logging
 import sys
+import threading
+import time
+import urllib.request
 from contextlib import asynccontextmanager
 
 from mcp.server.fastmcp import FastMCP
@@ -16,6 +19,8 @@ from mcp.server.transport_security import TransportSecuritySettings
 from .config import (
     VAULT_MCP_ALLOWED_HOSTS,
     VAULT_MCP_FORWARDED_ALLOW_IPS,
+    VAULT_MCP_HEARTBEAT_INTERVAL,
+    VAULT_MCP_HEARTBEAT_URL,
     VAULT_MCP_HOST,
     VAULT_MCP_PORT,
     VAULT_MCP_TOKEN,
@@ -27,6 +32,27 @@ logger = logging.getLogger(__name__)
 
 # Global frontmatter index instance
 frontmatter_index = FrontmatterIndex()
+
+
+def _heartbeat_ping(url: str) -> None:
+    """Send a single liveness GET. Split out from the loop so it is unit-testable."""
+    with urllib.request.urlopen(url, timeout=10) as resp:
+        resp.read()
+
+
+def _heartbeat_forever(url: str, interval: int) -> None:
+    """Ping ``url`` every ``interval`` seconds for the process lifetime.
+
+    Runs in a daemon thread started from main() -- NOT the per-request MCP lifespan,
+    which fires on every request and would spawn a heartbeat per session. Failures
+    are logged and swallowed so a flaky monitor can never take the server down.
+    """
+    while True:
+        try:
+            _heartbeat_ping(url)
+        except Exception as e:
+            logger.warning("Heartbeat failed: %s", e)
+        time.sleep(interval)
 
 
 @asynccontextmanager
@@ -259,6 +285,17 @@ def main():
     logger.info(f"Starting vault MCP server. Vault: {VAULT_PATH}")
     frontmatter_index.start()
     atexit.register(frontmatter_index.stop)
+
+    # Optional liveness heartbeat. Daemon thread tied to the process (not the
+    # per-request lifespan), started only when configured. No-op when unset.
+    if VAULT_MCP_HEARTBEAT_URL:
+        threading.Thread(
+            target=_heartbeat_forever,
+            args=(VAULT_MCP_HEARTBEAT_URL, VAULT_MCP_HEARTBEAT_INTERVAL),
+            daemon=True,
+            name="heartbeat",
+        ).start()
+        logger.info("Heartbeat enabled (interval: %ds)", VAULT_MCP_HEARTBEAT_INTERVAL)
 
     # Build the Starlette app with auth middleware and OAuth endpoints
     try:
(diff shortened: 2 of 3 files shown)


ProfessionalSeaweedDevourer (commented):
Thanks for this. The core design is right: daemon thread started from `main()` rather than the per-request lifespan (same reasoning as #28), `_heartbeat_ping` split out for testability, and failures logged and swallowed so a flaky monitor cannot take the server down. SSRF is a non-issue here since the URL is operator config, not request-derived input.

I reproduced every issue below empirically at this PR's head (f6c48c4), none are theoretical.

### Blocking

1. **A negative interval kills the heartbeat thread silently after one ping.** `time.sleep(interval)` sits outside the try/except in `_heartbeat_forever`, so `VAULT_MCP_HEARTBEAT_INTERVAL=-5` raises `ValueError: sleep length must be non-negative` after the first (successful) ping. Verified with a real thread: the thread dies, `threading.excepthook` fires (a stderr traceback that bypasses the logging setup), and the module logger records nothing. This is the one spot where "failures are logged and swallowed" does not hold: the monitor sees one ping and then permanent silence. `interval=0` fails the other way: verified 100+ pings in 150 ms, a busy loop against the monitor URL.

2. **A non-integer interval crashes the whole server at import time, even with the heartbeat disabled.** `int(os.environ.get("VAULT_MCP_HEARTBEAT_INTERVAL", "60"))` runs at module import, before `logging.basicConfig` in `main()`. Verified: `VAULT_MCP_HEARTBEAT_INTERVAL=60s` (and the empty string) kill `import obsidian_vault_mcp.config` with a raw traceback while `VAULT_MCP_HEARTBEAT_URL` is unset. A stale typo in an optional, off-by-default knob takes down all vault access. (It mirrors the existing `VAULT_MCP_PORT` pattern, granted, but the heartbeat being optional makes the blast radius harder to justify.)

3. **No startup validation of the URL, so permanent config errors masquerade as transient failures forever.** Verified: a typo'd scheme (`htp://...`) raises `URLError: unknown url type` and a missing scheme raises `ValueError`, both deterministically and before any network I/O. The loop catches them like a network blip and retries: three iterations produced three byte-identical warnings with the loop still going. At the default interval that is 1,440 identical lines per day, while the operator's monitor reports "no pings received" against a healthy server.

All three resolve with one small validation step in `main()` before starting the thread: scheme must be http(s), interval must be >= 1, and parse the int there (or defensively in config.py) with a clear error message. That matches the fail-loudly precedent of the `VAULT_PATH` check and the fail-closed auth-app construction. Narrowing the loop's catch to `(urllib.error.URLError, OSError, TimeoutError)` would then also stop it from absorbing future bugs in `_heartbeat_ping` itself. The scheme check additionally closes a small footgun I verified: a misconfigured `file:///...` "heartbeat" succeeds silently today and looks healthy.

### Important

4. **`test_loop_swallows_errors` can pass vacuously.** I mutation-tested this: adding `return` inside the `except` block (the exact "loop dies after first failure" regression this test exists to catch) still passes the test, because nothing asserts that the `KeyboardInterrupt` actually escaped. One-line fix, verified to kill that mutant and pass on the current code:

   ```python
   with pytest.raises(KeyboardInterrupt):
       server._heartbeat_forever("http://x", 1)
   ```

5. **README env-var table.** The table documents every other operator-facing variable, including optional ones; please add rows for `VAULT_MCP_HEARTBEAT_URL` and `VAULT_MCP_HEARTBEAT_INTERVAL`. Right now the config.py comment is the only documentation, so the feature is effectively undiscoverable.

6. **Warning context, with a caveat.** "Heartbeat failed: timed out" names neither the exception type nor the destination. But please do not fix it by logging the full URL: Healthchecks.io / Uptime Kuma push URLs are capability URLs whose path component is the secret, so logging them verbatim copies the credential into logs. Logging scheme+host plus `type(e).__name__` covers diagnosis (DNS typos, wrong host) without the leak. If omitting the URL from the "Heartbeat enabled" startup line was deliberate for the same reason, a short comment there would protect it from a future "helpful" change.

### Nice to have

- A consecutive-failure counter (warn on the 1st and every Nth, plus one INFO "recovered after N failures") would kill the log spam from finding 3 and give a post-mortem timeline for free.
- The `if VAULT_MCP_HEARTBEAT_URL:` wiring in `main()` is untested. Extracting a `_maybe_start_heartbeat()` helper would make both branches, plus the (url, interval) passing, testable with a monkeypatched `threading.Thread`. A `caplog` assertion in the loop test would also make its docstring's "is logged" claim true. Side note from writing my own repro tests: `_heartbeat_forever` has no termination mechanism at all, which is also what makes it awkward to exercise with real threads.
- In the `_heartbeat_forever` docstring, "fires on every request" vs "a heartbeat per session" mixes terms; the `lifespan` docstring says per request, which is also the stronger statement.

### What's good

- Thread placement in `main()` with the #28 rationale documented at the call site is exactly right, and `daemon=True` plus a named thread is the right shape.
- The ping/loop split is a clean testability seam, and the `KeyboardInterrupt` loop-escape trick (not an `Exception` subclass, so it pierces the catch) is the standard way to test a `while True` without threads or sleeps.
- The error suppression itself is the correct policy for transient monitor outages. The asks above only separate permanent configuration errors from the transient failures the catch is actually for.

jimprosser:
Useful feature and the architecture is right — off by default, a named daemon thread from `main()`, no vault data in the ping. A handful of small things before it can merge, mostly around failing safely on a bad value.

**Interval handling.** A non-int `VAULT_MCP_HEARTBEAT_INTERVAL` crashes the whole server at import (the `int()` runs at module load, even with the heartbeat disabled); a negative value kills the daemon thread after one ping because `time.sleep(interval)` sits outside the `try/except`; `0` busy-loops. Validate in `main()` before starting the thread — require an int ≥ 1 and fail with a clear message.

**URL handling.**
- No scheme validation: a typo'd/no-scheme URL is caught like a transient network blip and retried forever, spamming logs while the monitor shows nothing. Validate an `http(s)` scheme at startup.
- `urlopen()` follows redirects, which turns a configured monitor URL (or a compromised one) into a blind SSRF toward `localhost`/LAN/metadata endpoints. Disable redirect-following.
- The failure path logs `str(e)`, and a no-scheme error includes the full configured URL — push-monitor URLs are capability URLs where the secret *is* the path, so this leaks the secret into logs. Log scheme+host+exception type only, never the full URL.
- No response-size cap: a malicious or compromised endpoint can return an endless body and pin the thread/memory. Cap the read.

**Tests/docs.** `test_loop_swallows_errors` can pass vacuously (it would still pass if the loop died after the first failure) — wrap the call in `pytest.raises(KeyboardInterrupt)` so it actually catches that regression. And add `VAULT_MCP_HEARTBEAT_URL` / `VAULT_MCP_HEARTBEAT_INTERVAL` to the README env-var table so the feature is discoverable.

jimprosser:
Merged to `main` in 8046c34 (your authorship preserved). Hardening on top: heartbeat config is validated at startup and fails closed (http(s) URL with a host; positive-integer interval) with no import-time crash; the ping goes through a no-redirect opener with a capped body read; and failure logging emits only host + exception type, never the full URL — so the capability URL can't leak into logs even if the URL/interval env vars get swapped. I deliberately kept private/LAN targets working, since a self-hosted monitor on the LAN is the common case. Codex-confirmed merge-ready.

Closing since the squashed commit differs from this branch's head. Thanks for both #43 and #45 — exactly the kind of contributions I want.

Issue #50, opened 2026-06-11
Heavily-extended fork — which features would you welcome upstream?


mleitnercom:
Hi Jim — I maintain a long-running fork ([mleitnercom/obsidian-web-mcp](https://github.com/mleitnercom/obsidian-web-mcp)) that branched from the initial commit and has since added ~20 tools. Now that you're actively merging PRs, I'd like to contribute the non-overlapping parts back rather than let the two trees drift further — but coordinate first so I build them the way you want.

**Already converged independently** (you have these — I won't PR them): OAuth bypass/hardening, fail-closed `/oauth/authorize`, connector interop (serve at `/`, RFC 9728 metadata + challenge), DCR persistence, atomic-rename handling in the frontmatter index, configurable allowed-hosts, append/edit. Naming differs on my side (`VAULT_ALLOWED_HOSTS` vs your `VAULT_MCP_ALLOWED_HOSTS`; `vault_str_replace`/`vault_patch` vs your `vault_edit`) — I'd adopt your conventions for anything I upstream.

**Candidate features not in upstream** (happy to PR as separate, self-contained branches rebased on your `main`, with tests):

- **Tier 1 — pure filesystem, easiest to review:**
  - Canvas: `vault_canvas_read` / `add_node` / `add_edge` (JSON `.canvas` parsing, no plugin dependency)
  - Daily notes: `vault_daily_note_path` / `read` / `append`
- **Tier 2 — Obsidian Local REST API bridge:**
  - Templater: `vault_template_list` / `render` / `apply` (simple `{{var}}` renderer, not full Templater)
  - Dataview: `vault_dataview_query` (TABLE DQL)
- **Tier 3 — larger subsystems:**
  - Semantic search (`vault_semantic_search`, `vault_reindex`) — FAISS + fastembed, optional extra
  - JSONL audit pipeline (append-only per-mutation records, opt-in read auditing, `/health` status)
  - Signed direct-upload (`vault_request_upload_url` + `POST /upload/{id}`) and URL/file import (`vault_import_url` / `vault_import_file`)
  - Vault analytics (`vault_analytics_summary` / `findings`: frontmatter/link/tag/encoding hygiene)
  - PDF OCR sidecars (external OCR for scanned PDFs, persisted `.ocr.txt`)
- **Probably fork-specific (won't push unless you want a generic version):** `recurring_materialize` (recurring-task templating, tailored to my task system)

Questions:
1. Which of these do you actually want upstream?
[...]

PR #55, opened 2026-06-13
feat: daily-note tools (path, read, append) [Tier 1, #50]

mleitnercom:
Adds the daily-note half of Tier 1 (#50). Three pure-filesystem tools for a date-stamped daily note. Self-contained: no new routes, no dependencies, no subprocess, no outbound requests.

## What it does

A *daily note* is a single markdown file per calendar day (a common Obsidian workflow: journaling, a running log, capturing tasks). These tools let an MCP client work with "today's note" without computing the path or resending the body each turn.

| Tool | What it does |
|------|--------------|
| `vault_daily_note_path` | Resolve today's daily-note path from the server's local date. Does not read or create. |
| `vault_daily_note_read` | Read today's note. Returns an error payload (does **not** create it) when missing. |
| `vault_daily_note_append` | Append to today's note; creates it from the template when missing. Token-efficient daily logging. |

The path is `<FOLDER>/<date>.md`, built from three optional config knobs (all with safe defaults, all `strftime`):

- `VAULT_DAILY_NOTES_FOLDER` -- folder for daily notes; empty = vault root
- `VAULT_DAILY_NOTES_FORMAT` -- filename pattern, default `%Y-%m-%d`
- `VAULT_DAILY_NOTES_TEMPLATE` -- prepended once, when the note is first created

## Example

With `VAULT_DAILY_NOTES_FOLDER=Journal`, `VAULT_DAILY_NOTES_TEMPLATE="# %Y-%m-%d\n\n"`:

```
vault_daily_note_append(content="- 09:14 shipped the daily-note PR")
# creates Journal/2026-06-13.md as:
#   # 2026-06-13
#
#   - 09:14 shipped the daily-note PR
# a later append the same day adds to it; created=false
```

## How it fits the bar

- **No new attack surface.** Tools only; they run behind the existing bearer middleware. No new config that is risky (the knobs are path-shaping, defaults are safe), no new dependencies, no subprocess, no outbound.
- **Paths** go through `resolve_vault_path` (the resolved daily path is validated like any other).
- **Writes** reuse `vault_append`, i.e. the existing atomic write path; `append` creates-if-missing via that same path.
- **`read` does not create** -- a missing note returns an error payload rather than silently materialising a file.

## Tests (`tests/test_daily_notes.py`, 5 cases)

Path resolution honours folder/format; `read` on a missing note errors and does not create; `append` creates with the template then appends on the second call (`created` flips to false); `read` returns appended content; tools are registered and the `append` wrapper reaches the helper end to end.

Full suite green on Linux: **185 passed, 1 skipped** (180 baseline + 5 new). (Run on Linux/macOS -- `tests/test_oauth.py` uses `os.fchmod`, unavailable on Windows; unrelated to this PR.)

## CONTRIBUTING checklist

- [x] One self-contained change, rebased on `main` (not stacked)
- [x] Full test suite passes locally (`pytest`) -- 185 passed, 1 skipped on Linux
- [x] Auth: no new route/mount (tools only, behind existing bearer middleware); fails closed
- [x] Paths go through `resolve_vault_path`; resolved path re-validated
- [x] No `shell=True` on request input; subprocess env scrubbed -- N/A (no subprocess)
- [x] Outbound requests don't follow redirects to private ranges; size capped -- N/A (no outbound)
- [x] New config has safe defaults and is off/inert by default (empty folder = root, default format); validated by the existing path guard
- [x] Bridge/optional deps fail soft; heavy deps optional -- N/A (stdlib only)
- [x] Writes go through the atomic write path (`vault_append`)
- [x] README updated (three tools in the tool table + three env vars in the config table)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ Found a vulnerability? Please report it privately rather than opening a public i
 | `vault_list` | List directory contents with recursion depth, glob filtering, and file/dir toggles |
 | `vault_move` | Move or rename a file or directory within the vault |
 | `vault_delete` | Soft-delete a file by moving it to `.trash/` (requires explicit confirmation) |
+| `vault_daily_note_path` | Resolve today's daily-note path from the configured folder/format |
+| `vault_daily_note_read` | Read today's daily note; returns an error (does not create it) when missing |
+| `vault_daily_note_append` | Append to today's daily note, creating it from the template when missing |
 
 ## Prerequisites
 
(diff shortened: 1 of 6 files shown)


jimprosser:
Merged to `main` in eb0bf8d (your authorship preserved). I rebased onto main and resolved the conflict against the Canvas tools (#54), which landed first and touched the same import/registration blocks in `server.py`/`models.py` and the README tool table — both purely additive, so the resolution just keeps both sets. Full suite green afterward (199 passed, 1 skipped). Both halves of Tier 1 are now in. Thanks — clean PRs, exactly the shape I asked for. Closing since the rebased SHA differs from this branch's head.

PR #56, opened 2026-06-14
feat: append-only JSONL audit log for vault mutations [#50]

mleitnercom:
## What it does

Adds the optional append-only JSONL audit log we discussed on #50 -- the token-id-hash
shape you asked for, self-contained, off by default, fails closed. With
`VAULT_AUDIT_LOG_PATH` unset there is zero overhead; set it and every vault mutation
appends one JSON record.

| Env var | Default | Effect |
|---|---|---|
| `VAULT_AUDIT_LOG_PATH` | (none) | Append-only JSONL log of mutations. Empty = auditing off. |
| `VAULT_AUDIT_LOG_INCLUDE_READS` | `false` | Also record read/search ops (opt-in). |

Each record: `timestamp` (UTC), `token_id_hash` (SHA-256 of the bearer token -- the raw
token is never written), `client_id` (best-effort User-Agent hint), `operation`,
`target_path`, `size_before`/`size_after`, `checksum_before`/`checksum_after` (SHA-256),
`request_id`, `operation_status`, `error`.

`GET /health` (bearer-exempt) now returns an `audit` block: enabled, log path, includes_reads,
last write time, and 24h write-error / byte counts.

## Example

```jsonl
{"checksum_after":"9f86d0…","checksum_before":null,"client_id":"claude","error":null,"operation":"vault_write","operation_status":"success","request_id":"a1b2…","size_after":42,"size_before":null,"target_path":"notes/today.md","timestamp":"2026-06-14T18:30:00+00:00","token_id_hash":"5e88…"}
```

## How it fits the bar

- **No new attack surface that runs anything.** This is the audit half only -- it records,
  it never executes. (The post-write/exec hook from the same internal work is deliberately
  *not* here, per your call on #44.)
- **Principal threading is one ContextVar.** The bearer middleware records the principal
  (plus a request id + UA hint) in `context.py`; the tool layer reads it. The raw token
  stays in that context -- only its SHA-256 hash is ever persisted. No secrets in the log.
- **Fail-closed config.** `VAULT_AUDIT_LOG_PATH` is validated as writable at startup; an
  unwritable path aborts the server rather than silently dropping records. Off by default.
- **Audit can't break a write.** A record-write failure is logged and counted, never
  propagated to the tool result.
- **Paths via `resolve_vault_path`.** Snapshots resolve through the same guard, so a path
  escaping the vault is treated as absent, not read.
- **No new dependencies.** Wraps the existing tools through one `_run_audited` helper; tool
  schemas are unchanged.

The auth + server + health touch is the single self-contained change you said was fine here
("the plumbing is the feature"). It has been running on my own instance for ~a month.

## Tests

`tests/test_audit.py` (21 cases): record shape + all required fields, token hashed (raw
token asserted absent from the log), before/after size+checksum on overwrite and move, read
opt-in (off by default / on when enabled), error `operation_status`, audit-write failure
isolation (tool still succeeds), startup writability (file-parent rejected, mkdir-intermediates
allowed), path-traversal snapshot safety, and that every wrapped tool stays registered.

Full suite green on Linux: **220 passed, 1 skipped**. (`test_oauth.py` uses `os.fchmod`, so
the suite is validated on Linux, not Windows.)

## PR checklist

- [x] Auth: any new route/mount validated against the exempt set; fails closed
- [x] Paths go through resolve_vault_path; resolved path re-validated
- [x] No shell=True on request input; subprocess env scrubbed of secrets (n/a -- no subprocess)
- [x] Outbound requests don't follow redirects to private ranges; response size capped (n/a -- no outbound)
- [x] New config validated at startup, fails closed, off by default if risky
- [x] Bridge/optional deps fail soft when absent; heavy deps are optional extras (n/a -- no new deps)
- [x] Writes go through the atomic write path (n/a -- the log is append-only JSONL, not a vault write)
- [x] README updated for new env vars / tools / dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ All configuration is via environment variables:
 | `VAULT_DAILY_NOTES_TEMPLATE` | No | (none) | `strftime` template prepended when a daily note is first created |
 | `VAULT_MCP_HEARTBEAT_URL` | No | (none) | Optional push URL for an uptime monitor (Uptime Kuma, Healthchecks.io, ...). When set, a daemon thread GETs it on an interval. Must be `http(s)`; redirects are not followed and the URL is treated as a secret (never logged in full). Empty = disabled. |
 | `VAULT_MCP_HEARTBEAT_INTERVAL` | No | `60` | Seconds between heartbeat pings. Must be a positive integer; a bad value fails closed at startup. Only used when `VAULT_MCP_HEARTBEAT_URL` is set. |
-| `VAULT_AUDIT_LOG_PATH` | No | (none) | Append-only JSONL audit log of vault mutations. When set, every mutation appends one record; empty disables auditing. The raw bearer token is never written -- only its SHA-256 hash. Validated as writable at startup (an unwritable path **fails the server closed**). See [Audit logging](#audit-logging). |
+| `VAULT_AUDIT_LOG_PATH` | No | (none) | Append-only JSONL audit log of vault mutations. When set, every mutation appends one record; empty disables auditing. The raw bearer token is never written -- only its SHA-256 hash. Must resolve **outside** the vault and be writable; otherwise the server **fails closed** at startup. See [Audit logging](#audit-logging). |
 | `VAULT_AUDIT_LOG_INCLUDE_READS` | No | `false` | Also record read/search operations (`1`/`true`/`yes`/`on`). Off by default; mutations are always logged once the audit log is enabled. |
 
 Generate secrets with: `python -c "import secrets; print(secrets.token_hex(32))"`
@@ -152,12 +152,20 @@ raw token is never written), `client_id` (a best-effort User-Agent hint), `opera
 {"checksum_after":"9f86d0…","checksum_before":null,"client_id":"claude","error":null,"operation":"vault_write","operation_status":"success","request_id":"a1b2…","size_after":42,"size_before":null,"target_path":"notes/today.md","timestamp":"2026-06-14T18:30:00+00:00","token_id_hash":"5e88…"}
 ```
 
-The log path is validated at startup: if it is set but not writable, the server refuses to
-start rather than silently dropping records (fail-closed). A write failure at runtime is
-logged and counted but never alters the tool result -- the audit trail cannot break a write.
-The `GET /health` endpoint (no auth required) reports audit status under an `audit` key:
-whether it is enabled, the log path, whether reads are included, the last write time, and 24h
-write-error and byte counts.
+**Put the log outside the vault.** `VAULT_AUDIT_LOG_PATH` must resolve outside `VAULT_PATH`.
+A log inside the vault would be just another file the vault tools can reach, so an
+authenticated caller could overwrite it (`vault_write`) or move it (`vault_delete`) and
+defeat the append-only premise. The server validates this at startup and **refuses to start
+(fail-closed)** if the path is not writable or resolves inside the vault.
+
+**Threat model — the log is best-effort at runtime, not tamper-evident.** A write failure
+at runtime is logged to the server log but never alters the tool result (the audit trail
+must not be able to break a write), so a record can be dropped silently; the server log is
+the only signal. Batch mutations emit one record per file with that file's own status, so a
+partial failure is never recorded as a whole-batch success. The unauthenticated `GET /health`
+endpoint reports only `{"status": "ok", "audit": {"enabled": <bool>}}` — it deliberately does
+not expose the log path or write counters (which would leak host filesystem layout and a
+vault-activity side-channel to anonymous callers over the tunnel).
 
 ## Connecting to Claude
 
--- a/README.md
+++ b/README.md
@@ -131,8 +131,33 @@ All configuration is via environment variables:
 | `VAULT_DAILY_NOTES_TEMPLATE` | No | (none) | `strftime` template prepended when a daily note is first created |
 | `VAULT_MCP_HEARTBEAT_URL` | No | (none) | Optional push URL for an uptime monitor (Uptime Kuma, Healthchecks.io, ...). When set, a daemon thread GETs it on an interval. Must be `http(s)`; redirects are not followed and the URL is treated as a secret (never logged in full). Empty = disabled. |
 | `VAULT_MCP_HEARTBEAT_INTERVAL` | No | `60` | Seconds between heartbeat pings. Must be a positive integer; a bad value fails closed at startup. Only used when `VAULT_MCP_HEARTBEAT_URL` is set. |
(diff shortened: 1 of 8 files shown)


jimprosser (changes requested):
Thanks for this — it's well-scoped and the design is sound: off by default with zero overhead when unset, no new deps, no exec surface (the audit half only, per the call on #44), raw token never persisted (only SHA-256), JSON serialization that prevents log injection, and snapshot paths routed through `resolve_vault_path`. The PR checklist is accurate.

I ran a second, independent automated review pass in addition to my own. Four things to resolve before merge, then I'm happy to take it.

## Requested changes

### 1. Unauthenticated `/health` discloses audit internals
`/health` is bearer-exempt (`auth._AUTH_EXEMPT_PATHS`) and the new `audit` block returns the absolute `log_path`, `last_write_at`, `write_errors_count_24h`, and `bytes_written_24h`. This server is meant to run on a public Cloudflare Tunnel, so that hands any anonymous caller the host filesystem layout (the path typically leaks the OS username) plus a vault-activity side-channel (when, and how much, the vault is being written). Please keep the unauthenticated `/health` to liveness only — e.g. `{"status": "ok", "audit": {"enabled": <bool>}}` — and gate the path/counters behind auth (or drop them).

### 2. The "append-only" log is tamperable when it lives inside the vault
`audit_log_path()` accepts any path, and startup only validates writability. If an operator sets `VAULT_AUDIT_LOG_PATH` to a non-dotfile path under `VAULT_PATH`, any authenticated writer can overwrite the log via `vault_write` or relocate it via `vault_delete` — `resolve_vault_path` only blocks traversal and dotfiles, so a same-vault audit log is just another writable file. That defeats the integrity premise of the feature. Please reject an audit path that resolves inside `VAULT_PATH` at startup (fail closed), or exclude the resolved audit-log path from the vault tools, and document the constraint in the README.

### 3. Batch operations record incorrect status, not just thin metadata
In `_run_audited`, status is derived only from a top-level `"error"` key:
```python
status = "error" if "error" in parsed else "success"
```
But `vault_batch_frontmatter_update` reports per-file failures inside `results` (e.g. `{"results": [{"path": "a.md", "updated": true}, {"path": "b.md", "error": "..."}]}`). A batch in which some files fail is therefore logged as `operation_status: "success"`. On top of that, `target_path` comes back as a list, which `snapshot_path` ignores (non-string), so successful batch writes also get null before/after size+checksum. For an audit trail, "reports success on a partial failure" is the part that matters. Please emit one record per file, or a structured multi-target record with per-target status and snapshots.

### 4. The feature's load-bearing path isn't exercised by the tests
All of `test_audit.py` binds the request context by hand (`context.set_request_context(...)`), bypassing `BearerAuthMiddleware` — so the one thing the whole feature depends on, the principal/`client_id`/`request_id` reaching the tool layer, is never actually tested. I verified it does work on the currently pinned `mcp 1.26.0` + `starlette 0.52.1` (FastMCP calls sync tools inline, and the ContextVar set before `call_next` survives the middleware's task boundary). But that's a version-pinned guarantee: a future bump that runs sync tools in a threadpool, or reworks the middleware, would silently null every `token_id_hash` with the suite still green. Please add one HTTP-level regression test that drives an authenticated request through the real app and asserts the record's `token_id_hash`, UA-derived `client_id`, and middleware-generated `request_id`.

Here's a minimal version of the propagation test I used (proves the middleware → sync-handler boundary, which is the only boundary in question given FastMCP's inline call):

```python
"""The principal set by BearerAuthMiddleware must reach SYNC downstream code -- the path
the audit log depends on and that test_audit.py does not cover (it binds context in-frame)."""
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route
from starlette.testclient import TestClient

from obsidian_vault_mcp import auth as auth_module
from obsidian_vault_mcp.context import current_request_context


def test_principal_propagates_to_sync_handler(monkeypatch):
    monkeypatch.setattr(auth_module, "VAULT_MCP_TOKEN", "secret-token")

    def sync_probe(request):  # sync on purpose: mirrors how FastMCP runs vault_* tools
        ctx = current_request_context()
        return JSONResponse({"principal": ctx.get("principal"), "client": ctx.get("client"),
                             "request_id": ctx.get("request_id")})

    app = Starlette(routes=[Route("/probe", sync_probe)])
    app.add_middleware(auth_module.BearerAuthMiddleware)
    client = TestClient(app)

    r = client.get("/probe", headers={"Authorization": "Bearer secret-token",
                                       "User-Agent": "probe/1"})
    assert r.status_code == 200, r.text
    body = r.json()
    assert body["principal"] == "secret-token"   # null here => token_id_hash null in prod
    assert body["client"] == "probe/1"
    assert body["request_id"]


def test_context_clean_outside_request():
    assert current_request_context() == {}
```
An HTTP-level test through the real MCP tool call (asserting the fields land in the actual log file) would be even better.

## Non-blocking
- `vault_edit(dry_run=True)` is logged as a successful mutation though nothing is written — consider tagging `dry_run` or skipping.
- `vault_daily_note_append` passes no path to the wrapper, so it gets no before-snapshot.
- Runtime is fail-open by design (a write failure is swallowed so the audit can't break a write). That's a reasonable call for a non-tamper-evident log, but please state it in the README threat-model so operators know the log is best-effort at runtime and the only signal of dropped records is the (currently unauthenticated) `/health` error counter.

Appreciate the care that went into this — items 1–4 are all small, and #2 is the one I'd most want closed since it's the integrity property the feature is selling.

mleitnercom:
Thanks for the thorough pass — all four are addressed in `21bc0a4`, plus the non-blocking items. Full suite green on Linux (278 passed, 1 skipped).

**1. Unauthenticated `/health` disclosure.** `/health` is now liveness-only: `{"status": "ok", "audit": {"enabled": <bool>}}`. The log path and write counters are gone, and I removed the 24h-counter machinery that only existed to feed that detailed payload.

**2. In-vault tamperability (the integrity one).** `audit_path_inside_vault()` rejects a `VAULT_AUDIT_LOG_PATH` that resolves inside `VAULT_PATH` at startup (fail-closed), next to the writability check. The README now states the constraint and the reasoning.

**3. Batch status + null snapshots.** Batch mutations now go through `_run_audited_batch`, which emits **one record per file** with that file's own `operation_status` and real before/after snapshots. A batch where some `results[]` entries fail is recorded per-file (the failing file gets `operation_status: "error"`), never as a single whole-batch success — and the list `target_path` no longer collapses to null snapshots.

**4. Untested load-bearing path.** Added `tests/test_audit_propagation.py` — your propagation test (authenticated request → `BearerAuthMiddleware` → sync handler, asserting `principal` / `client_id` / `request_id` land), plus a `/health` liveness assertion. Thanks for writing it out.

**Non-blocking:** `vault_edit(dry_run=True)` is no longer recorded as a mutation; `vault_daily_note_append` now passes its path so it gets a before-snapshot; and the README documents the runtime fail-open / best-effort threat model (a dropped record only shows in the server log) so operators know what the log does and doesn't guarantee.

mleitnercom:
Rebased onto `main` after the extension seam (#57) landed. The audit `/health` block and the `serve()` startup validation now sit alongside the new `build_app(extensions=())` / extension hooks — no interaction between them, and the extension-route auth guard is untouched. Full suite green on Linux (294 passed, 1 skipped).

PR #57, opened 2026-06-14
feat: extension seam — serve(extensions=()) + FrontmatterIndex change-listener/rebuild API

jimprosser:
## What it does

Adds an opt-in **extension seam** so a downstream deployment can add tools, HTTP routes, and index hooks **without forking `server.py`**. `main()` now calls a new `serve()` with no extensions, so the stock server is behaviorally unchanged.

A custom deployment provides its own entry point:

```python
from obsidian_vault_mcp.server import serve
from obsidian_vault_mcp.extensions import Extension

class MyExtension(Extension):
    def register_tools(self, mcp): ...
    def before_indexes_start(self, frontmatter_index): ...
    def after_indexes_start(self, frontmatter_index): ...
    def register_routes(self, app): ...
    def shutdown(self): ...

def main():
    serve([MyExtension()])
```

## Changes

- **`extensions.py`** (new) — `Extension` base class with five no-op lifecycle hooks. Override only what you need.
- **`server.py`** — `main()` → `serve(extensions=())`; `build_app(extensions=())`. Hooks fire in a defined order: `register_tools` → `before_indexes_start` → (index starts) → `after_indexes_start` → `register_routes`; `shutdown` via `atexit` (LIFO, so it runs before the frontmatter index stops). Extension routes are added **before** the bearer middleware, and `build_app()` **fails closed** if an extension route collides with an auth-exempt path (`/health`, `/oauth/*`, `/.well-known/*`) — an extension can't accidentally expose an unauthenticated endpoint.
- **`frontmatter_index.py`** — adds `add_change_listener(callback)` and `rebuild()`. Backward-compatible: no behavior change with zero listeners; listener exceptions are logged and swallowed.
- **`tests/test_extensions.py`** — 11 tests: hook ordering, extension-route auth, auth-exempt collision rejected, generator input handled, `atexit` ordering, and the listener/`rebuild()` API.
- **README** — adds an "Extending the server" section.

## Tests

Full suite green: **261 passed, 1 skipped**.

## PR checklist

- [x] One self-contained change, rebased on main
- [x] New behavior covered by tests; the wiring is tested, not just helpers
- [x] Auth: extension routes are bearer-protected; collision with an auth-exempt path fails closed
- [x] No new unauthenticated surface when no extensions are loaded (stock server unchanged)
- [x] No new dependencies
- [x] No secrets/tokens in logs
- [x] README updated for the new extension API

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,14 +308,23 @@ def main():
     serve([MyExtension()])
 ```
 
-The hooks run in the order above. Two guarantees worth knowing:
+The hooks run in the order above.
 
-- **Extension routes are authenticated.** They're registered before the bearer-auth
-  middleware, so they require the bearer token like every other route. `build_app()`
-  **fails closed** if an extension route collides with an auth-exempt path (`/health`,
-  `/oauth/*`, `/.well-known/*`) so you can't accidentally expose an unauthenticated
-  endpoint. (The method-only off-root `GET/HEAD /` probe isn't path-checkable — don't
-  register there.)
+> **Trust model — read this.** Extensions are **fully-trusted, in-process code** that you
+> choose to load. An extension runs with the server's full privileges: it can read the
+> bearer token and OAuth secrets, read/write your vault, and mutate any route. This is
+> **not a sandbox** — only load extensions you wrote or trust, like any dependency.
+
+Two things worth knowing:
+
+- **Extension routes are authenticated, with a footgun guard.** Routes are registered
+  before the bearer-auth middleware, so they require the bearer token like every other
+  route. As a guardrail against honest mistakes, `build_app()` **fails closed** if an
+  extension adds a route that would cover an auth-exempt path (`/health`, `/oauth/*`,
+  `/.well-known/*`, or the off-root `GET/HEAD /` probe) — including via a wildcard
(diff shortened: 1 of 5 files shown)


mleitnercom:
As you know I maintain a heavily-extended fork (Canvas and Daily Notes came out of it; audit is #56), so this is exactly the seam I'd re-platform onto — the additive surface (tools, routes, index listeners) covers most of what I carry.

The one thing it doesn't reach is the cross-cutting layer. My audit threads the authenticated principal from `BearerAuthMiddleware` to the tool layer (the `token_id_hash` in #56); rate-limiting is similar. Those are middleware concerns, so today they'd stay core rather than become extensions.

Would a narrow, read-only hook there be in scope — something like `on_authenticated_request(principal, request)` fired right after the bearer check, so an extension can stash the principal in its own contextvar without touching auth itself? That would let cross-cutting features (audit, rate-limit, per-principal metrics) live as extensions too. Completely fine if you'd rather keep the auth path core and out of the extension contract — your call; the additive seam alone is already a big deal for me.

`add_change_listener` / `rebuild()` and `before/after_indexes_start` are the parts I'd use most.

Issue #53, opened 2026-06-13
vault_search / read_file: an in-vault hardlink to an outside file leaks its contents (symlinks are refused, hardlinks are not)


ProfessionalSeaweedDevourer:
## Summary

A hardlink created inside the vault that points at a file **outside** `VAULT_PATH` leaks the outside file's contents through both `vault.read_file` and `vault_search`. Symlinks are correctly refused in both paths; hardlinks are not.

This is a pre-existing, vault-wide confidentiality gap (not specific to one tool). Filing it separately from #39, where the new search context re-read shares the same blind spot and I deliberately left it out of scope.

## Why hardlinks slip through while symlinks don't

A symlink is resolved by `realpath`, points outside, and is rejected by the containment check in `resolve_vault_path` (and, on the #39 search re-read, by `O_NOFOLLOW`). A hardlink is a second real directory entry for the same inode: `realpath("<vault>/hl.md")` returns `<vault>/hl.md` (in-vault), and the inode carries no back-reference to its other (outside) name, so there is no path-based way to detect it.

## Reproduction (macOS, against current behavior)

A hardlink and a symlink inside the vault, both pointing at `outside/secret.txt`:

```
=== vault.read_file ===
  read_file("hardlink.md"): returned content, leaked_outside=True
  read_file("symlink.md"):  REFUSED (ValueError: Path resolves outside the vault root)

=== vault_search context ===
  search "hardlink.md": match_context = 'ctx a\nthe OUTSIDE-SECRET token\nctx b'  (leaked)
  search "symlink.md":  no match (rg doesn't follow symlinks; re-read refuses it anyway)
```

So a symlink out of the vault is refused, but a hardlink to the same target is read and returned.

## Threat model

Requires a **local** actor able to create a hardlink inside the vault directory, on the same filesystem as the target. The remote MCP surface (a connected client, or adversarial note content via prompt injection) cannot create hardlinks, so it does not reach this. On Linux, `fs.protected_hardlinks=1` (default on modern distros) blocks hardlinking to files the actor doesn't own / can't write, narrowing it further; macOS has no equivalent.

Low severity, but flagging because the project's bar is explicitly about what earns its attack surface on a tunnel-exposed server holding a personal vault.

## Possible directions (your call)

- Reject files with `st_nlink > 1` on read. Simple, but also rejects legitimate hardlinks and doesn't actually prove the other link is outside the vault.
[...]

PR #59, opened 2026-06-15
feat: vault analytics — hygiene summary + findings (read-only) [#50]

mleitnercom:
## What it does

Adds two read-only, pure-filesystem vault analytics tools (the "hygiene findings" you marked welcome on #50):

- `vault_analytics_summary` — compact vault-hygiene summary: counts plus a few examples of missing frontmatter, missing required fields, broken `[[wikilinks]]` (classified repairable path-mismatch / missing-target / ambiguous), near-duplicate tag variants, and markdown files that are not valid UTF-8.
- `vault_analytics_findings` — the detailed list for a single category.

Both take an optional `path_prefix`, walk from `resolve_vault_path` (vault root when empty), skip `EXCLUDED_DIRS` and symlinks, and serialize through `serialization.dumps`.

## How it fits the bar

- **Read-only, no new surface:** no plugin, no network, no subprocess, no new config, and no new dependency (reuses the existing `python-frontmatter`). `readOnlyHint` annotations.
- **Paths through `resolve_vault_path`** (traversal/dotfile/null-byte guard); a bad `path_prefix` returns the standard `{"error", "path_prefix"}` payload, never a stack trace.
- **No fork infrastructure** — plain `@mcp.tool` wrappers registered exactly like the Canvas/Daily tools; no rate-limit, no audit wrapper.

## Tests

`tests/test_analytics.py` (13 tests): summary/findings happy paths over a temp vault (missing frontmatter, broken wikilinks, tag variants, non-UTF-8 files), wikilink classification edge cases (source-relative, repairable, missing, ambiguous, links-in-frontmatter, excluded dirs), abuse cases (`path_prefix` of `..`, `../escape`, and a dotfile dir all return clean errors), and the `@mcp.tool` wiring seam. Full suite green on Linux: **279 passed, 1 skipped** (`test_oauth.py` uses `os.fchmod`, so the suite is validated on Linux).

## PR checklist

- [x] Auth: any new route/mount validated against the exempt set; fails closed (n/a — no routes)
- [x] Paths go through resolve_vault_path; resolved path re-validated
- [x] No shell=True on request input; subprocess env scrubbed of secrets (n/a — no subprocess)
- [x] Outbound requests don't follow redirects to private ranges; response size capped (n/a — no outbound)
- [x] New config validated at startup, fails closed, off by default if risky (n/a — no new config)
- [x] Bridge/optional deps fail soft when absent; heavy deps are optional extras (n/a — no new deps)
- [x] Writes go through the atomic write path (n/a — read-only)
- [x] README updated for new env vars / tools / dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ Found a vulnerability? Please report it privately rather than opening a public i
 | `vault_daily_note_path` | Resolve today's daily-note path from the configured folder/format |
 | `vault_daily_note_read` | Read today's daily note; returns an error (does not create it) when missing |
 | `vault_daily_note_append` | Append to today's daily note, creating it from the template when missing |
+| `vault_analytics_summary` | Compact vault-hygiene summary: counts and examples of missing frontmatter, broken wikilinks, near-duplicate tag variants, and non-UTF-8 files |
+| `vault_analytics_findings` | Detailed findings for one analytics category (`frontmatter_missing`, `required_frontmatter_missing`, `broken_wikilinks`, `suspicious_tag_variants`, `encoding_issues`, `oversized_files`) |
 
 ## Prerequisites
 
--- a/src/obsidian_vault_mcp/models.py
+++ b/src/obsidian_vault_mcp/models.py
@@ -446,3 +446,60 @@ class VaultDailyNoteAppendInput(BaseModel):
         description="Content to append to today's daily note (the note is created from the template if missing)",
         max_length=MAX_CONTENT_SIZE,
     )
+
+
+class VaultAnalyticsSummaryInput(BaseModel):
+    """Build a compact analytics summary for a vault path."""
+
+    model_config = ConfigDict(str_strip_whitespace=True, extra="forbid")
+
+    path_prefix: str | None = Field(
+        default=None,
+        description="Optional folder prefix to restrict the analysis",
+        max_length=500,
+    )
+    required_frontmatter: list[str] | None = Field(
+        default=None,
+        description="Optional required frontmatter fields to validate",
+        max_length=20,
+    )
+    max_examples: int = Field(
+        default=3,
+        ge=1,
+        le=20,
+        description="Maximum example findings to include per category",
+    )
+
+
+class VaultAnalyticsFindingsInput(BaseModel):
+    """Return detailed findings for one analytics category."""
+
+    model_config = ConfigDict(str_strip_whitespace=True, extra="forbid")
+
+    category: Literal[
+        "frontmatter_missing",
+        "required_frontmatter_missing",
+        "broken_wikilinks",
+        "suspicious_tag_variants",
+        "encoding_issues",
+        "oversized_files",
+    ] = Field(
+        ...,
+        description="Analytics finding category to return",
+    )
+    path_prefix: str | None = Field(
+        default=None,
+        description="Optional folder prefix to restrict the analysis",
+        max_length=500,
+    )
+    required_frontmatter: list[str] | None = Field(
+        default=None,
+        description="Optional required frontmatter fields to validate",
+        max_length=20,
+    )
+    max_results: int = Field(
+        default=50,
+        ge=1,
+        le=200,
+        description="Maximum number of findings to return",
+    )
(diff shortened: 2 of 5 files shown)


jimprosser:
Read-only and clean on the bar — no routes, no auth/`/health` surface, no new deps, the registered tools are wired-tested, and there's abuse coverage for traversal, dotfiles, and unknown categories. Two things before it lands:

1. **The whole-vault read is unbounded.** `_load_posts` pulls every file fully into memory with no per-file cap, so one pathological multi-hundred-MB note (or a very large vault) spikes memory on a server reachable over the tunnel. Add a size cap: skip or truncate a file over the limit and surface it as a finding, consistent with the 1MB instinct on the write path.
2. **Each post carries a second full copy of the file** in a `text` field (plus an unused `name`) that nothing downstream reads — only `body`/`frontmatter`/`path` are used. Drop them; for a tool that already walks the whole vault, a duplicate copy of every file is pure overhead.

The in-vault-hardlink read isn't yours to fix here — it's the same `st_nlink` gap the rest of the read path on `main` has, tracked on #53/#39. Fix those two and I'll merge.

mleitnercom:
Addressed both:

1. **Per-file read cap.** `_load_posts` now reads at most `config.MAX_CONTENT_SIZE` (1 MB) per file via a bounded `open().read(cap)` — top-of-file frontmatter still parses, so a >1 MB note is analysed, not skipped wholesale, but one pathological file (or a huge vault) can't spike memory.
2. **Surfaced as a finding.** New `oversized_files` category: summary count + examples, and `vault_analytics_findings("oversized_files")` lists `{path, size_bytes, limit_bytes}` — a capped file is visible, not silent.
3. **Dropped the per-post `text` duplicate and the unused `name`.** Only `body` / `frontmatter` / `path` are read downstream.

The in-vault-hardlink (`st_nlink`) read is the shared `main` gap on #53/#39, left out here as you said. Rebased onto current `main`. Full suite green on Linux: **362 passed, 1 skipped**.

jimprosser:
One gap before this lands: `oversized_files` is in the `category_map` in `vault_analytics_findings` and advertised in the README, but it's **not in the `Literal` in `VaultAnalyticsFindingsInput`** in `models.py`. The server.py wrapper instantiates that model before calling the tool, so `vault_analytics_findings("oversized_files")` through the MCP endpoint fails Pydantic validation — the category the summary now surfaces can't actually be queried.

The tests for it (`test_oversized_file_is_capped_and_flagged`) call the analytics module directly, bypassing the model, so they pass without catching this.

Two-line fix:

**`models.py`** — add `"oversized_files"` to the Literal:
```python
    category: Literal[
        "frontmatter_missing",
        "required_frontmatter_missing",
        "broken_wikilinks",
        "suspicious_tag_variants",
        "encoding_issues",
        "oversized_files",   # add this
    ] = Field(...)
```

**`server.py`** — update the tool description annotation to include `oversized_files` in the category list (it currently stops at `encoding_issues`).

Everything else looks good — the cap logic, the findings surfacing, the tests, the wikilink classification edge cases. Just needs this one fix to make the advertised capability actually reachable through the MCP endpoint.

mleitnercom:
Fixed. Added `oversized_files` to the `Literal` in `VaultAnalyticsFindingsInput` and to the `vault_analytics_findings` description in `server.py`, so the advertised category is now reachable through the MCP endpoint.

Also added `test_oversized_files_category_reachable_through_model` — it drives the category through the model and the server wrapper (not just the analytics module, which is what let the original test miss this), and asserts an unknown category still raises. Good catch.

Rebased onto current `main`. Full suite green on Linux: **372 passed, 1 skipped**.

jimprosser:
Re-reviewed after the fix. Verified independently, not on the test claim:

- `oversized_files` is now in the `Literal` in `VaultAnalyticsFindingsInput` (models.py) and in the `server.py` tool description — the category the summary surfaces is now reachable through the MCP endpoint.
- `test_oversized_files_category_reachable_through_model` is the right guard: it builds the model with `category="oversized_files"` and drives `server.vault_analytics_findings(...)` through the wrapper (the path the original test bypassed), and asserts an unknown category still raises. This would have caught the original miss.
- Rebase held: per-file 1MB read cap in `_load_posts`, `oversized_files` finding with `{path, size_bytes, limit_bytes}`, and the duplicate `text`/`name` fields dropped.

Cloned the branch and ran the suite locally: 372 passed, 1 skipped. Merging.

PR #61, opened 2026-06-16
feat: vault_write_binary — write allowed binary files (image/PDF) from base64

mleitnercom:
## What it does

A first-class write path for binary attachments, which the server currently lacks: `vault_write` / `write_file_atomic` are text-only, so there is no way to place an image or PDF in the vault through MCP. One new tool.

| Tool | Description |
|---|---|
| `vault_write_binary` | Write an allowed binary file (image/PDF) to the vault from base64 content; media-type allowlist + size cap, atomic write |

Plumbing: `write_bytes_atomic` in `vault.py` (the binary counterpart of the existing `write_file_atomic` — tempfile + `os.replace`), a conservative media-type/extension allowlist, and `config.MAX_BINARY_SIZE` (10 MB).

## Example

```
vault_write_binary(path="attachments/diagram.png", data="<base64>", media_type="image/png")
# -> {"path": "attachments/diagram.png", "created": true, "size": 20431, "media_type": "image/png"}
```

## How it earns its attack surface

Against the CONTRIBUTING bar ("does it earn its attack surface on a network-reachable server holding private notes"):

- **No new outbound surface, no new route, no subprocess, no new dependency.** A local write behind the same bearer middleware as every other tool. The client supplies the bytes; the server fetches nothing — contrast a URL-import path, which is SSRF surface and is deliberately *not* proposed here.
- **The allowlist is the boundary.** Not arbitrary-file write: `media_type` must be in a conservative allowlist (PNG/JPEG/WebP/GIF/SVG/PDF) **and** the target extension must match it. Unknown type or path/extension mismatch is rejected before any bytes are written.
- **Vault-confined.** Target goes through `resolve_vault_path` (traversal/dotfile/escape rejected) before the write.
- **Bounded.** Size capped by `config.MAX_BINARY_SIZE` (10 MB); the input model also caps the base64 length, so an oversized payload is rejected *before* it is decoded into memory. Strict base64 (`validate=True`).
- **Atomic.** `write_bytes_atomic` writes to a tempfile then `os.replace`s — Obsidian Sync never sees a partial file, same guarantee as the text path. Overwrite is opt-in (default `false`).

## Tests

`tests/test_write_binary.py` — happy path, invalid base64, unsupported media_type, extension/media-type mismatch, path traversal (`..`), overwrite guard, size cap, model validation. Wiring exercised through the tool function, not just the helper.

Full suite green on Linux: **274 passed, 1 skipped** (skip = pre-existing optional-dep test).

## PR checklist

- [x] Full test suite passes locally (`pytest`) — 274 passed on Linux
- [x] Auth: no new route/mount; tool sits behind the existing bearer middleware
- [x] Paths go through `resolve_vault_path`; resolved path used for the write
- [x] No `shell=True` on request input; no subprocess
- [x] Outbound requests: N/A (no outbound request; client supplies the bytes)
- [x] New config (`MAX_BINARY_SIZE`) is a static safety limit; allowlist conservative; overwrite off by default
- [x] Bridge/optional deps: N/A; no new/heavy deps (stdlib `base64` only)
- [x] Writes go through the atomic write path (`write_bytes_atomic`)
- [x] README updated (Tools table)

Note: the media-type allowlist is intentionally a fixed conservative set — happy to make it operator-extensible (`VAULT_*` env) in a follow-up if you'd want that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Found a vulnerability? Please report it privately rather than opening a public i
 | `vault_read` | Read a file, returning content, metadata, and parsed YAML frontmatter |
 | `vault_batch_read` | Read multiple files in one call; handles missing files gracefully |
 | `vault_write` | Write a file with optional frontmatter merging; creates parent dirs |
+| `vault_write_binary` | Write an allowed binary file (image/PDF) to the vault from base64 content; enforces a media-type allowlist (declared type/extension, not byte-sniffed) and size cap, writes atomically |
 | `vault_edit` | Patch a file with ordered exact text replacements (token-efficient partial edits); supports dry-run diff previews |
 | `vault_append` | Append content to the end of a file without resending the existing body; creates the file when missing |
 | `vault_batch_frontmatter_update` | Update YAML frontmatter fields on multiple files without touching body content |
--- a/src/obsidian_vault_mcp/config.py
+++ b/src/obsidian_vault_mcp/config.py
@@ -154,6 +154,7 @@ VAULT_AUDIT_LOG_INCLUDE_READS = os.environ.get(
 
 # Safety limits
 MAX_CONTENT_SIZE = 1_000_000  # 1MB max write size
+MAX_BINARY_SIZE = 10_000_000  # 10MB max binary write size (images/PDFs run larger than text)
 MAX_BATCH_SIZE = 20           # Max files per batch operation
 MAX_SEARCH_RESULTS = 50       # Max results per search
 DEFAULT_SEARCH_RESULTS = 20
--- a/src/obsidian_vault_mcp/models.py
+++ b/src/obsidian_vault_mcp/models.py
@@ -8,6 +8,7 @@ from .config import (
     CONTEXT_LINES,
     DEFAULT_SEARCH_RESULTS,
     MAX_BATCH_SIZE,
+    MAX_BINARY_SIZE,
     MAX_CONTENT_SIZE,
     MAX_LIST_DEPTH,
     MAX_SEARCH_RESULTS,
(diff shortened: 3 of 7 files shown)


jimprosser:
Clean PR overall — true allowlist, atomic `write_bytes_atomic` (same-dir `mkstemp` + `os.replace`), the size cap is enforced at the model layer *before* the base64 decode, `validate=True`, and the abuse tests are there.

**One blocker: drop `image/svg+xml` from the default allowlist.** SVG is the one entry that isn't passive — it carries `<script>`/`onload` — and since `_validate_binary_target` gates only on the declared `media_type` + extension and never sniffs bytes, that's an arbitrary-active-content write into a vault that may be synced or rendered in a preview surface. PNG/JPEG/WEBP/GIF/PDF are inert; SVG isn't, and "image/PDF" in the title doesn't promise it. Pull SVG and I'll merge. If you have a real need for it, say so and we'll do a sanitize-or-reject path — default-off is the baseline.

Two smaller things while you're in there (not blockers):

- `overwrite=false` is check-then-replace: `exists()` in `vault_write_binary`, then `write_bytes_atomic` re-checks, then `os.replace` — which still clobbers if another writer creates the file in between. For a true no-clobber, use an `O_EXCL`/`os.link`-style atomic create when `overwrite` is false.
- The allowlist verifies the *name*, not the *bytes* — a caller can write any bytes under a `.png`/`.pdf`. Fine for a single-user vault (you only fool yourself), but worth a line in the docstring/README so it isn't over-trusted, and PDF-can-carry-active-content is worth a conscious note since it's in the set.

mleitnercom:
Addressed the review:

- **Blocker — dropped `image/svg+xml`** from the default allowlist. Agreed it's the one non-inert entry, and since validation is by declared `media_type` + extension (no byte-sniffing) it'd be an arbitrary-active-content write. PNG/JPEG/WEBP/GIF/PDF only now, with a test asserting SVG is rejected.
- **No-clobber is now atomic.** `write_bytes_atomic(overwrite=False)` uses `os.link` (fails if the target exists) instead of `exists()`-then-`os.replace`, closing the check-then-write race.
- **Documented the name-not-bytes trust boundary** (and the PDF-can-carry-active-content note) in the `vault_write_binary` docstring and the README row.

Rebased onto current `main` (after the audit / write-listener merges). Full suite green on Linux: **357 passed, 1 skipped**.

Issue #58, opened 2026-06-15
Extension seam has no write-side hook — add a write listener


jjsmackay:
#57 added a clean way to extend the server without forking — `serve(extensions=())` plus `register_tools` / `register_routes` and, on the index, `add_change_listener`. That covers reading and reacting to *indexed file content*. There's no equivalent on the **write** side: an extension can't observe the built-in mutation tools.

### The gap

For an extension that needs to react to a vault mutation *as an operation* — a git commit with a meaningful message, an audit/activity log, a webhook/notify — neither existing hook fits:

- `register_tools` only *adds* tools. It can't observe `vault_write` / `vault_append` / `vault_edit` / `vault_batch_frontmatter_update` / `vault_move` / `vault_delete` without re-registering shadow tools that wrap them, which is fragile and fights the existing names.
- `add_change_listener` is the right shape but answers a different question. It's watcher-driven, so it gives `(abs_path, exists)` after a debounce with:
  - **no operation verb** — created/updated/moved/deleted all collapse to `exists: bool`; a rename arrives as two uncorrelated events;
  - **no provenance** — a filesystem watcher can't distinguish a tool-initiated write from an external edit (Obsidian Sync, `git pull`), which is exactly the distinction an attributed-commit or audit use case needs;
  - **per-path, not per-operation** — a batch update is N callbacks, not one;
  - **`.md` only** — the handler filters by suffix, so attachments/canvas/etc. are invisible;
  - **a `rebuild()` blind spot** — listeners fire only from `_flush_pending`; the reconcile path swaps the index without notifying.

### Proposed shape

A small in-process registry, deliberately a mirror of `add_change_listener` — no-op with zero listeners (byte-identical stock server), exceptions logged and swallowed:

```python
# core
def register_write_listener(cb): ...      # cb(operation: str, paths: list[str])
def fire_write(operation, paths): ...      # called at each write tool's success path
```

Wired at the success path of the six mutation tools, with the semantics they already know about: created-vs-updated distinguished, `vault_move` passing both paths as one `"moved"`, batch passing only the successfully-written paths, no fire on a dry-run / no-change `vault_edit`, delete firing only on success.

[...]

PR #62, opened 2026-06-21
feat: write-listener seam — register_write_listener + public fire_write (#58)

jjsmackay:
Implements #58 — the write-side analogue of `FrontmatterIndex.add_change_listener` (#57).

## Why

The extension seam (#57) lets you observe *indexed content* via `add_change_listener`, but that's watcher-driven: it collapses create/update/move/delete into `(abs_path, exists)`, can't distinguish a tool write from an external edit (Sync, `git pull`), fires per-path not per-operation, and is `.md`-only. An extension that needs a vault mutation *as an operation* — a provenance-aware git commit, an audit log, a webhook — has nothing to subscribe to. `register_tools` only *adds* tools; it can't observe `vault_write`/`vault_append`/`vault_edit`/`vault_batch_frontmatter_update`/`vault_move`/`vault_delete`.

This is the redesign of the declined #44: instead of a `shell=True` hook in core, core gains only a no-op callback list and **all side effects move downstream** into the (already fully-trusted) extension.

## What

New `write_events.py`:
- `register_write_listener(cb)` — subscribe; `cb(operation, paths)`.
- `fire_write(operation, paths)` — **public** publish side, so an extension that writes on its own path (e.g. attachment bytes that never pass through a core tool) can join the same stream.

Wired at the success path of the six core mutation tools. Semantics:
- created-vs-updated distinguished;
- `vault_move` fires one `"moved"` with `[source, destination]`;
- a batch fires **once** with only the successfully-written paths;
- no fire on a dry-run or no-change `vault_edit`, a no-change `vault_append`, or any error;
- `vault_delete` fires only on a confirmed success.

Zero listeners ⇒ byte-identical no-op; a listener's exception is logged and swallowed so it can't break the write or starve the other listeners.

Design discussion (public publish side, and keeping the principal/before-state out of this seam) is in #58 — thanks @mleitnercom for shaping it from the audit-extension side.

## Tests

21 new tests (`test_write_events.py` + `test_write_listener_integration.py`) exercising the tools, not just the helper, with the negative cases (dry-run, no-change, error, unconfirmed delete). No CI on this repo — green local run is the gate:

```
287 passed, 1 skipped
```

## Checklist

- [x] Auth: no new route/mount
- [x] Paths: no new path handling (rides the existing tools' resolved paths)
- [x] No `shell=True`; no subprocess at all (the point — side effects stay downstream)
- [x] Outbound requests: none
- [x] No secrets/tokens/capability URLs in logs (logs only operation + vault-relative paths, mirroring the change-listener log)
- [x] No new config; zero-listener no-op is the off-by-default
- [x] No new dependencies
- [x] Writes still go through the atomic path (fire is *after* the existing write)
- [x] README updated (extension-seam section)

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,7 @@ you need) and run the server with `serve([YourExtension()])` from your own entry
 ```python
 from obsidian_vault_mcp.server import serve
 from obsidian_vault_mcp.extensions import Extension
+from obsidian_vault_mcp.write_events import register_write_listener
 
 
 class MyExtension(Extension):
@@ -290,6 +291,8 @@ class MyExtension(Extension):
     def before_indexes_start(self, frontmatter_index):
         # e.g. attach a change listener so no change is missed once the index starts
         frontmatter_index.add_change_listener(self._on_change)
+        # ...or react to a mutation as an operation (see write-event seam below)
+        register_write_listener(self._on_write)  # _on_write(operation, paths)
 
     def after_indexes_start(self, frontmatter_index):
         # e.g. start a periodic reconcile loop now that the index is live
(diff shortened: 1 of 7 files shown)


mleitnercom:
+1 from the downstream side. The **public `fire_write`** is the part that makes this usable for me: my import / attachment-write extensions write bytes on their own path (never through the six core tools), so without a callable publish side an audit or sync listener would never see those writes — with it, they all land in one stream.

The `(operation, paths)`-only scope is the right call too: keeping the principal and before/after out of this seam is what keeps it a clean no-op primitive. Would adopt as soon as it lands.

Issue #60, opened 2026-06-15
Extension seam has no read-side content-extractor hook


mleitnercom:
#57 added a clean read/index side (`add_change_listener` / `rebuild`) and #58 proposes the write side (`register_write_listener`). The remaining gap is the **read path itself**: an extension can add new tools, but it can't influence what `read_file` / `vault_read` returns for an existing file.

### The gap

Some content is derivable at read time, not stored: scanned-PDF text (OCR), a transcript for an audio attachment, a rendered preview for a binary format. Today an extension can only expose this as a *separate* tool (e.g. `vault_pdf_ocr(path)`) — which works, but isn't transparent: a client that calls `vault_read` on a scanned PDF gets empty text and has to know to call the other tool.

### Proposed shape (narrow, read-only)

A small registry mirroring `add_change_listener` — the host calls registered extractors inside `read_file` only when the built-in extraction yields nothing (or for a file type the host doesn't handle):

```python
def register_content_extractor(cb): ...   # cb(path: str, default_text: str) -> str | None
# called inside read_file when default extraction is empty/unsupported;
# first non-None wins; exceptions logged and swallowed; zero extractors => byte-identical
```

### Security / threat model

Same bar as #44/#57 — "does a feature earn its attack surface" — applied vector by vector:

- **Zero surface on a stock server.** With no extractor registered, `read_file` is byte-identical and the registry is a no-op callback list. The tunnel-exposed-private-vault default is unchanged. (Same shape as `add_change_listener` and the #58 write-listener.)
- **No new capability for an extension.** Extensions are already fully-trusted, in-process code per #57 — they can read `VAULT_MCP_TOKEN`, the whole filesystem, any route. A read-extractor grants nothing a loaded extension can't already do; a malicious extension needs no hook to exfiltrate. This does not widen the trust boundary you already documented ("not a sandbox; only load ones you trust").
- **Downstream of auth, not a new entry point.** The hook fires *inside* `read_file`, only reached after the bearer middleware and `resolve_vault_path`. It adds no route and no unauthenticated path — so there's none of the route-collision concern extension routes have. It can only act on a file the caller already requested and is already authorized to read.
- **Read-only, scoped to the requested file.** The signature returns text for the one path being read; it can't redirect the read to another file, escalate, or write. The worst a buggy/hostile extractor can do is return wrong text *for that file* — bounded to content the caller was already getting, squarely inside the "trusted code you chose to load" risk #57 already scopes out.
[...]

PR #63, opened 2026-06-22
feat: read-side content-extractor seam (register_content_extractor)

mleitnercom:
## What it does

The read-side mirror of the write listener (#62) and the index change-listener (#57). `content_extractors.register_content_extractor(cb)` lets an extension supply text for a file the host can't read itself — OCR for a scanned PDF, a transcript for audio, a rendered preview — transparently through `read_file`, instead of as a separate must-know-about tool. Implements #60.

`cb(path, default_text) -> str | None` is consulted inside `read_file` **only** when the built-in extraction is empty or unsupported; the first non-None result wins; exceptions are logged and swallowed; off by default.

## Example

```python
from obsidian_vault_mcp.content_extractors import register_content_extractor

register_content_extractor(lambda path, default_text: ocr_pdf(path) or None)
# vault_read("scans/contract.pdf") now returns the OCR text transparently
```

## How it fits the bar

- **Zero surface on a stock server.** With no extractor registered, `read_file` is byte-identical: a normal read is unchanged, and a non-UTF-8 / binary file re-raises exactly as before (the hook returns `None`, the original error propagates). Same shape as `add_change_listener` and the #62 write listener.
- **Downstream of auth, no new route.** The hook fires inside `read_file`, only reached after the bearer middleware and `resolve_vault_path`. No route, no unauthenticated path, no auth-exempt surface.
- **Read-only, scoped to the requested file.** It returns text for the one path being read; it can't redirect to another file, escalate, or write. The worst a buggy/hostile extractor does is return wrong text *for that one file* — bounded to content the caller already requested and is authorized to read (documented in the docstring + README, per your note).
- **No core shell-out (the #44 line).** Core gains only a no-op callback list; the OCR binary / model / subprocess lives entirely in the extension.
- **Failure-isolated.** Consulted only when the built-in extraction is empty/unsupported; first non-None wins; an extractor's exception is logged and swallowed so a flaky extractor can't break a read.
- **No new config, no new dependency.**

## Tests

`tests/test_content_extractors.py` — exercised through the real `read_file` (the seam, not just the helper): byte-identical no-op for a text file, binary-still-raises with no extractor, fill on empty + unsupported, first-non-None wins, extractor exception swallowed, not-consulted-for-normal-text, and the empty-registry helper. Full suite green on Linux: **356 passed, 1 skipped**.

## PR checklist

- [x] One self-contained change, rebased on main (not stacked on another PR) — branched off current `main`
- [x] Full test suite passes locally (`pytest`) — 356 passed on Linux
- [x] New/abuse inputs covered by tests; the seam is tested through `read_file`, not just the helper
- [x] Auth: no new route/mount; the hook is downstream of the bearer middleware, inside `read_file`
- [x] Paths go through `resolve_vault_path` (rides `read_file`'s already-resolved path); no new path handling
- [x] No `shell=True` on request input; no subprocess (lives in the extension)
- [x] Outbound requests: none
- [x] No secrets/tokens/capability URLs in logs (logs only `Content extractor error for <vault-relative path>`)
- [x] New config validated at startup, fails closed, off by default if risky — no new config; zero-extractor no-op is the off-by-default
- [x] Bridge/optional deps fail soft when absent; heavy deps are optional extras — N/A; no new deps (the OCR dep lives in the downstream extension)
- [x] Writes go through the atomic write path — N/A (read-only hook)
- [x] README updated for new env vars / tools / dependencies — extension-seam section

Implements #60.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,15 @@ Two things worth knowing:
   side, `fire_write(operation, paths)`, is public so an extension that writes on its own path
   can join the same stream. Use it for a provenance-aware commit, an audit log, or a webhook;
   a listener's exception is logged and swallowed.
+- **Content extractors fill the read side.** `content_extractors.register_content_extractor(cb)`
+  lets an extension supply text for a file the host can't read itself — OCR for a scanned PDF,
+  a transcript, a rendered preview. `cb(path, default_text) -> str | None` is consulted inside
+  `read_file` **only** when the built-in extraction is empty or unsupported; the first non-None
+  result wins, exceptions are logged and swallowed, and with none registered `read_file` is
+  byte-identical to stock. Note that a registered extractor changes what `read_file` returns for
+  that path, so a buggy one returns wrong text *for that one file* — bounded to the file the
+  caller already requested. The OCR/model/subprocess lives entirely in the extension; core only
+  gains a no-op callback list.
 
 ## VPS Setup With Cloudflare Origin TLS + Caddy Reverse Proxy
 
--- a/src/obsidian_vault_mcp/content_extractors.py
+++ b/src/obsidian_vault_mcp/content_extractors.py
@@ -1,0 +1,58 @@
+"""In-process read-side content-extractor seam: the read mirror of the write-event seam
+(``write_events``) and the index change-listener (#57).
+
+Lets an extension supply text for a file the host cannot read itself -- OCR for a scanned
+PDF, a transcript for an audio attachment, a rendered preview -- transparently through
+``read_file``, instead of as a separate must-know-about tool. Core stays a no-op callback
+list: with zero extractors registered ``read_file`` is byte-identical to the stock server,
+the OCR binary / model / subprocess lives entirely downstream in the (fully-trusted)
+extension, and an extractor's exception is logged and swallowed.
+
+    from obsidian_vault_mcp.content_extractors import register_content_extractor
+
+    register_content_extractor(lambda path, default_text: ocr(path) or None)
+
+NOTE: a registered extractor changes what ``read_file`` returns for that path. The hook
+fires only when the built-in extraction is empty or unsupported, the first non-None result
+wins, and a buggy extractor therefore returns wrong text *for that one file* -- bounded to
+the file the caller already requested and is authorized to read, but real, so only load
+extractors you trust (the #57 extension trust model).
+"""
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+# Registered at startup (before serving), consulted during request handling.
+_content_extractors: list = []
+
+
+def register_content_extractor(callback) -> None:
+    """Register a ``callback(path: str, default_text: str) -> str | None``.
+
+    Consulted by ``read_file`` ONLY when the built-in extraction is empty or unsupported
+    (e.g. a non-UTF-8 / binary file). ``path`` is vault-relative; ``default_text`` is
+    whatever the built-in read produced (an empty string when unsupported). Return the
+    extracted text, or ``None`` to decline -- the next extractor, then the host's default
+    behaviour, applies. With none registered this seam is a true no-op on the stock server.
+    Exceptions raised by an extractor are logged and swallowed, never propagated.
+    """
+    _content_extractors.append(callback)
+
+
+def apply_content_extractors(path: str, default_text: str) -> str | None:
+    """Return the first non-None extractor result for ``path``, or ``None`` if none apply.
+
+    Called inside ``read_file`` at the empty/unsupported branch; a no-op returning ``None``
+    when no extractor is registered. An extractor's exception is logged and swallowed so a
+    flaky extractor can't break a read.
+    """
+    for extractor in _content_extractors:
+        try:
+            result = extractor(path, default_text)
+        except Exception:
+            logger.warning("Content extractor error for %s", path)
+            continue
+        if result is not None:
+            return result
+    return None
--- a/src/obsidian_vault_mcp/vault.py
+++ b/src/obsidian_vault_mcp/vault.py
@@ -8,6 +8,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 from . import config
+from .content_extractors import apply_content_extractors
 
 
 def resolve_vault_path(relative_path: str) -> Path:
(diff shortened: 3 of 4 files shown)


jimprosser (changes requested):
Thanks. The seam is small and matches the shape of the write-listener and index-listener hooks, and it's a true no-op with nothing registered; I verified both.

It can't merge as is, though, because `read_file` is not only the read tool's read. It is also the read half of `vault_edit`, `vault_append`, `vault_batch_frontmatter_update`, and `vault_write(merge_frontmatter=True)`, each of which reads, transforms, and writes back. With a perfectly correct OCR extractor registered, `vault_edit` on a scanned PDF reports `changed: true` and replaces the PDF on disk with the OCR text; append and the frontmatter batch do the same. Without the extractor all four fail safely. An empty `.md` gets the extractor's text persisted on the next append. That is a data-loss path, not a wrong-text-for-one-file bound, and my note on #60 missed it too.

Please move the extractor call out of `vault.read_file` and into the read tools (`vault_read`, `vault_batch_read`), or gate it behind a keyword that defaults off and only the read tools pass. Add the negative test: extractor registered, `vault_edit` and `vault_append` on a binary file still error and leave the bytes untouched. I'd also drop the empty-file trigger unless you have a concrete need for it.

Separately, please link the extension that will call `register_content_extractor`. I'd rather freeze the callback signature against real use than a hypothetical one.

mleitnercom:
Reworked per your review, on top of eecfe6d (merged forward rather than force-pushed, so the diff against main is the whole change).

- The extractor call is gone from the default `read_file` path: `read_file(..., extract=True)` is passed only by `vault_read` and `vault_batch_read`. `vault_edit`, `vault_append`, `vault_batch_frontmatter_update` and `vault_write(merge_frontmatter=True)` keep failing on a non-UTF-8 file.
- Negative test through the registered MCP tools: it first proves the extractor returns text via `vault_read`, then shows each write-back tool errors and leaves the bytes untouched. Against the previous revision all four replace the binary on disk.
- Empty-file trigger dropped.
- Signature is now `callback(relative_path, path)`, with `path` host-resolved and past the hardlink check, because that is what the real consumer needs: https://github.com/mleitnercom/obsidian-vault-mcp-ext/blob/main/src/obsidian_vault_mcp_ext/ocr/extractor.py (docs: https://github.com/mleitnercom/obsidian-vault-mcp-ext/blob/main/docs/ocr.md).
- Validated on Linux with ripgrep, tesseract and poppler installed: 436 passed, 0 skipped. The extension's real-OCR test renders a page with pdftoppm and reads it with tesseract through `vault_read`, then checks `vault_append` leaves the image untouched.

jimprosser (changes requested):
Thanks for the fast turnaround. All four points from the first review are done, and I checked them against the code. Every `read_file` caller in `src/` defaults to `extract=False`, and only `vault_read` and `vault_batch_read` opt in. With `extract` flipped to default-on locally, all four cases of `test_write_back_tool_leaves_the_binary_untouched` fail along with the end-to-end test, so the negative tests catch what they claim to. The OCR extractor in your ext repo takes exactly `extract(relative_path, path)` and needs the absolute path for its subprocess argv, which settles the signature. Suite is 437 passed, 0 skipped here with ripgrep on PATH. Two small things and it merges.

1. `tests/_live_server.py` builds the child's environment from `**os.environ`. Anyone with `VAULT_AUDIT_LOG_PATH` or `VAULT_MCP_HEARTBEAT_URL` exported in their shell gets test writes appended to their real audit log and pings fired at their real heartbeat URL, and `VAULT_MCP_HOST=0.0.0.0` puts a server with a published token on the network. `VAULT_MCP_PATH=/mcp` breaks the helper outright, since the client endpoint is hard-coded to `/`. Build the env from scratch instead: `PATH`, the `VAULT_*` values the helper already sets, and `HOME` pointed at `tmp_path`, which also keeps the child away from the real `oauth_clients.json`. Add a `proc.wait()` after the `kill()` fallback so the child is reaped. The same file ships in #80, so fixing it here fixes it once.

2. Mark extracted content in the response. `vault_read("scan.pdf")` now returns OCR text in `content` with nothing to say it isn't the file. A model asked to fix a typo will try `vault_edit`, get the safe decode error, and fall back to `vault_write("scan.pdf", <text>)`, which has no extension check and replaces the PDF. That write is possible on main today, but nothing on main hands a model the text that makes it likely. Set `metadata["extracted"] = True` in `read_file` when an extractor supplied the content; both read tools already pass `metadata` through, so `tools/read.py` doesn't change. One test through the registered tool.

Not blocking: `apply_content_extractors` accepts any non-None result, so `lambda *_: 123` comes back as `"content": 123`. Treat a non-`str` as a decline and log it. And "never has to resolve or re-validate it" in the docstring and README promises more than the host can deliver, because the hardlink check and the extractor's own open are separate syscalls. "Already resolved and checked by the host" is accurate.

mleitnercom:
Both points done, plus the two non-blocking ones, on top of 34b1e67.

1. `tests/_live_server.py` builds the child's environment from scratch: `PATH` and a few process basics, the `VAULT_*` values the helper owns, `HOME` in `tmp_path`. Nothing is inherited, so an exported `VAULT_AUDIT_LOG_PATH`, `VAULT_MCP_HEARTBEAT_URL`, `VAULT_MCP_HOST` or `VAULT_MCP_PATH` cannot reach the child, and the child stays away from the real `oauth_clients.json`. `proc.wait()` after the `kill()` fallback. Two of the new tests are behavioural rather than dict checks, so they also run against the old helper and fail there: with `VAULT_AUDIT_LOG_PATH` exported the old child writes into it, and with `VAULT_MCP_PATH=/mcp` the endpoint answers 405 at `/`.

2. `read_file` sets `metadata["extracted"] = True` when an extractor supplied the content, and only then, so an ordinary read is byte-identical. Tested through both registered read tools, plus the absence of the key on a normal note.

Non-blocking: a non-`str` result is a decline and is logged (a lone `lambda *_: 123` now leaves the read failing instead of returning `"content": 123`). The docstring and README say "already resolved and checked by the host" instead of promising the extractor never has to re-validate.

Server run with ripgrep, tesseract and poppler present: 461 passed, 0 skipped.

jimprosser (approved):
Both done, and the two optional ones as well. I checked rather than read: 461 passed here with ripgrep, and with the old version of each file put back under your new tests the right ones fail every time, including the audit-path and `/mcp` helper tests you described. Pinning `VAULT_MCP_HOST` and `VAULT_MCP_PATH` in the child was a good addition.

Merging. Two things I'll handle on main rather than send you round again. The docstring and README still say "first non-None result wins", which is now "first string". And a second reviewer noticed the in-process tests will write to an exported `VAULT_AUDIT_LOG_PATH`; that is `conftest.py` on main (18 records from two existing test files when I tried it), not this PR.

Thanks for the fast, careful turnaround. #80 whenever you're ready.

PR #64, opened 2026-06-25
feat: signed direct binary upload (vault_request_upload_url + POST /upload/{id})

mleitnercom:
## What it does

A client-pushes-the-bytes path for binary files too large for `vault_write_binary`'s base64 argument: the client requests a short-lived HMAC-signed URL, then `POST`s the bytes straight to `/upload/{id}`. The server fetches nothing (no SSRF surface). Builds on `write_bytes_atomic` from #61.

| Tool | Description |
|---|---|
| `vault_request_upload_url` | Return a short-lived HMAC-signed upload URL; the client then `POST`s the bytes to `/upload/{id}` |

## Example

```
vault_request_upload_url(path="attachments/scan.pdf", media_type="application/pdf", max_size_bytes=4000000)
# -> { upload_url, upload_id, expires_at, curl, ... }
# then: curl -X POST -H "Content-Type: application/pdf" --data-binary @scan.pdf "<upload_url>"
```

## How it fits the bar

- **`POST /upload/{id}` is bearer-exempt because the HMAC signature in the URL IS the authorization.** Before any byte is written, `commit_direct_upload` verifies server-side: exact `expires` match against the stored metadata, not-expired, constant-time HMAC (`hmac.compare_digest`), single-use (`completed_at`), `Content-Type` matches the requested `media_type`, both size caps (per-URL `max_size_bytes` and the server `MAX_BINARY_SIZE`), the optional `expected_sha256`, and the target against the binary allowlist + `resolve_vault_path`.
- **Auth surface is narrow and explicit.** `auth.py` gains a single `/upload/` exempt *prefix* (it previously had only exact paths); `build_app`'s footgun guard now also rejects any extension route that would land under `/upload/`. No other path changes. Every check fails closed.
- **No outbound request, no subprocess, no new dependency.** The client supplies the bytes; the server only writes them via the existing atomic path. (Contrast a URL-import path, which is SSRF surface and deliberately not here.)
- **Secret + staging.** `VAULT_UPLOAD_URL_SECRET` is the HMAC key (falls back to `VAULT_MCP_TOKEN`). Staging holds only per-upload `metadata.json`, OUTSIDE the vault, with stale entries swept. The signed URL is a capability URL and is never logged (only `upload_id` + the rejection reason).
- **Write goes through the atomic path** (`write_bytes_atomic`: tempfile + `os.replace`, or `os.link` no-clobber when `overwrite=false`).

## Tests

`tests/test_signed_upload.py` (12) — happy path; bad signature; expired; single-use replay; oversize; media-type mismatch; checksum mismatch; SVG rejected; oversize-max; unknown id; plus wiring: the route is registered in `build_app`, and `POST /upload/{id}` is bearer-exempt end-to-end via TestClient (a normal route still `401`s; `/upload/` reaches the handler; a bad signature is `403`, never `401`).

Full suite green on Linux: **369 passed, 1 skipped**.

## PR checklist

- [x] One self-contained change, rebased on main (prepared on top of #61, rebased onto main now that #61 has landed)
- [x] Full test suite passes locally (`pytest`) — 369 passed on Linux
- [x] New/abuse inputs covered by tests; the route + auth-exempt wiring is tested, not just the helper
- [x] Auth: the one new route is validated against the exempt set — it is *intentionally* bearer-exempt (HMAC is the auth), the exemption is a narrow `/upload/` prefix, and `build_app` fails closed if an extension route would cover it
- [x] Paths go through `resolve_vault_path` (via `_validate_binary_target`) at request AND commit time
- [x] No `shell=True` on request input; no subprocess
- [x] Outbound requests: none (the client uploads; the server fetches nothing)
- [x] No secrets/tokens/capability URLs in logs (only `upload_id` + the rejection reason)
- [x] New config validated/fails closed, off by default if risky — HMAC secret falls back to `VAULT_MCP_TOKEN`; URLs are short-lived + single-use
- [x] Bridge/optional deps: N/A; no new dependencies (stdlib `hmac`/`hashlib`)
- [x] Writes go through the atomic write path (`write_bytes_atomic`)
- [x] README updated (Tools table)

Builds on #61 (now merged).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Found a vulnerability? Please report it privately rather than opening a public i
 | `vault_batch_read` | Read multiple files in one call; handles missing files gracefully |
 | `vault_write` | Write a file with optional frontmatter merging; creates parent dirs |
 | `vault_write_binary` | Write an allowed binary file (image/PDF) to the vault from base64 content; enforces a media-type allowlist (declared type/extension, not byte-sniffed) and size cap, writes atomically |
+| `vault_request_upload_url` | Return a short-lived, HMAC-signed URL for a direct binary upload; the client then `POST`s the bytes to `/upload/{id}` (server fetches nothing). For files larger than `vault_write_binary`'s base64 path. Same allowlist + size cap |
 | `vault_edit` | Patch a file with ordered exact text replacements (token-efficient partial edits); supports dry-run diff previews |
 | `vault_append` | Append content to the end of a file without resending the existing body; creates the file when missing |
 | `vault_batch_frontmatter_update` | Update YAML frontmatter fields on multiple files without touching body content |
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -29,6 +29,12 @@ _AUTH_EXEMPT_METHOD_PATHS = (
     {("GET", "/"), ("HEAD", "/")} if config.VAULT_MCP_PATH != "/" else set()
 )
 
+# Path PREFIXES exempt from bearer auth. The signed direct-upload endpoint
+# POST /upload/{id} carries its OWN authorization: an HMAC signature in the URL that
+# commit_direct_upload verifies (exact-expiry + single-use + constant-time compared). It is
+# therefore bearer-exempt by design. Kept narrow and explicit so it can't widen by accident.
+_AUTH_EXEMPT_PATH_PREFIXES = ("/upload/",)
+
 
 def _www_authenticate(request: Request, error: str) -> str:
     """RFC 9728 challenge header pointing clients at the protected-resource metadata.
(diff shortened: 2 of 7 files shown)


jimprosser:
Thanks for the careful write-up; the commit-side checks are solid and I verified them through the real route. I'm closing this rather than iterating, for two reasons.

First, as submitted the route does not work: `server.py` never imports the `config` module, so `direct_upload` raises `NameError` on every request that carries a `Content-Length` header, which is every curl and browser upload, including the example in the PR body. The wiring test builds its own Starlette app with its own handler, so it never exercises the handler `build_app()` installs. Please test through `build_app()` next time.

Second, the design does not clear the bar for a new unauthenticated route on this server. With the exemption in place, `VAULT_MCP_PATH=/upload/mcp/x` passes `validate_config()` and serves the whole vault without a token (verified: `vault_read` returned the note body with no `Authorization` header). An extension route at `/upload/x/y` slips past the new guard. The body is buffered in full before the id or signature is checked, so anyone on the internet can push 50 MB per request at the server. The stale sweep `rmtree`s any old directory under the staging path. Uploads bypass the audit log and write listeners. And uvicorn's access log records the signed URL. Each is fixable, but together they are a redesign, and the cap is still `MAX_BINARY_SIZE`, so this does not let anyone write a file `vault_write_binary` (#61) cannot already write.

If you want to pursue a large-file path, the shape I would consider: reserve `/upload` in `_validate_mcp_path`, validate the grant before reading any body bytes, stream to a tempfile under a hard cap, raise the cap so it does something #61 cannot, audit the commit, and cover it end to end through `build_app()`.

PR #66, opened 2026-07-10
feat: opt-in Cloudflare Access auth mode

thomastoye:
## What & why

Adds an **opt-in Cloudflare Access** authentication mode. Instead of exposing the built-in password login + static bearer token on the public internet, you can run the server behind a Cloudflare Tunnel with **Cloudflare Access + Managed OAuth**: Cloudflare authenticates the user at its edge (per-identity SSO / one-time-PIN) and injects a signed `Cf-Access-Jwt-Assertion` header on every request; the server's only job is to verify it.

This *reduces* the server's attack surface rather than widening it: it removes the public password endpoint (a brute-force target), gives per-identity, dashboard-revocable access (no redeploy), and lets Cloudflare handle rate-limiting / bot protection. It's strictly opt-in and off by default.

## Activation (both-or-neither)

| Setting | Description |
|---------|-------------|
| `VAULT_MCP_CF_ACCESS_TEAM_DOMAIN` | Cloudflare team domain, e.g. `myteam.cloudflareaccess.com` (bare name or full URL accepted). |
| `VAULT_MCP_CF_ACCESS_AUD` | Access application **AUD** tag. |

Set **both** to enable; if either is unset the mode is off and behavior is **byte-identical to today** (the app's own OAuth + static bearer). Install the deps with `pip install '.[cloudflare-access]'`.

## Behavior when on

- Every request except `/health` must carry a valid `Cf-Access-Jwt-Assertion`. The server verifies **signature (RS256, against Cloudflare's published keys at `/cdn-cgi/access/certs`), expiry, audience, and issuer**, and requires the `exp`/`iss`/`aud` claims. Missing / malformed / expired / wrong-signature / wrong-audience / wrong-issuer / unreachable-JWKS → `401`. **Fails closed** — never falls back to unauthenticated access or the static bearer.
- The app's own OAuth routes and `/.well-known/oauth-*` discovery are **not mounted**, and `VAULT_MCP_TOKEN` is **not accepted**. Cloudflare is the sole auth authority (no second, competing auth surface).
- The audit log records the caller's **identity** (`email` claim, or `sub` for service tokens) instead of a shared-token hash.

## Implementation notes

- New self-contained `src/obsidian_vault_mcp/cf_access.py` (config gate, JWKS verification, `CloudflareAccessMiddleware`). `build_app()` branches once on `cf_access_enabled()`; the off path is unchanged.
- **New dependency:** `pyjwt` + `cryptography`, shipped as the optional extra `cloudflare-access` (not in the base install) and imported lazily, so the default install and the off path never touch them. CF mode on + libs missing → **fail closed at startup** (`sys.exit(1)` with an install hint).
- Team domain is **validated at startup** (fail closed on a malformed value); connectivity stays a non-fatal warm-up warning so a transient Cloudflare blip can't wedge boot.
- Transport host-allowlist and the `/health` probe are unchanged. Alg-confusion (`alg=none`, HS256-with-public-key) is covered by the RS256 pin and regression tests.

## Security note on the JWKS fetch

The JWKS is fetched via PyJWT's `PyJWKClient`, which restricts schemes to `http(s)` and uses a request timeout, but does not itself block redirects or cap the response body. This is intentionally left as-is: the fetch URL is **operator config** (`VAULT_MCP_CF_ACCESS_TEAM_DOMAIN`), not request-influenced, and is always `https://<team>.cloudflareaccess.com/...`, so it isn't an attacker-controlled/SSRF vector. Happy to add a no-redirect + size-capped opener (mirroring the heartbeat) if you'd prefer belt-and-suspenders.

## Testing

Full suite green locally (no CI on the repo): **415 passed** (373 pre-existing + 42 new). New tests mint real RS256 tokens with an in-test RSA key and exercise the real `jwt.decode` path (only the JWKS *fetch* is stubbed). Coverage includes both modes and the abuse cases: missing / malformed / expired / wrong-aud / wrong-iss / bad-signature / `alg=none` / HS256-confusion → 401; OAuth routes not served; static bearer not accepted; `/health` open; and mode-off still mounts OAuth (non-breaking).

## Docs

README gets a "Cloudflare Access mode (opt-in)" section (what it does, the two settings, the install extra, a Cloudflare Managed OAuth setup walkthrough, and a "run behind Cloudflare" caveat about token replay / policy bypass), and both settings are added to the master configuration table.

## Checklist

- [x] One self-contained change, rebased on `main` (not stacked on another PR)
- [x] Full test suite passes locally (`pytest`) — 415 passed
- [x] Auth: no new routes; the new middleware exempts only `/health` and fails closed; `VAULT_MCP_PATH` still can't collide with an exempt path
- [x] Paths go through `resolve_vault_path` — N/A (no new path handling)
- [x] No `shell=True` on request input; subprocess env scrubbed — N/A (no subprocess)
- [x] Outbound requests: the JWKS fetch is `https`/scheme-restricted + timed out; URL is operator-config, not request-influenced (redirect/size-cap rationale above)
- [x] No secrets/tokens/capability URLs in logs (only exception type + the public team domain)
- [x] New config validated at startup, fails closed, off by default (both-or-neither; malformed domain and missing deps both fail closed)
- [x] Optional/heavy deps are an optional extra (`.[cloudflare-access]`), imported lazily; CF mode on + missing → fail closed at startup
- [x] Writes go through the atomic write path — N/A (no writes)
- [x] README updated for new env vars / dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,104 @@ This is a server that provides network access to your personal notes. Security i
 
 **Your vault is never exposed directly to the internet.** The recommended deployment uses a Cloudflare Tunnel -- an outbound-only encrypted connection. Your machine opens no inbound ports, and the server itself binds to loopback (`127.0.0.1`) by default. The login above is the authentication boundary; you can additionally layer Cloudflare Access (SSO, device posture, IP restrictions) on top for defense in depth.
 
+### Cloudflare Access mode (opt-in)
+
+Instead of the built-in password login + static bearer token, you can hand
+authentication entirely to **Cloudflare Access**. Cloudflare authenticates the user at
+its edge (per-identity SSO / one-time-PIN) and, on every request it forwards through the
+tunnel, injects a signed `Cf-Access-Jwt-Assertion` header carrying the caller's identity.
+The server's only job is to verify that header.
+
+**Enable it** by setting **both** of these (both-or-neither — if either is unset the mode
+is off and the server behaves exactly as it does by default):
+
+| Setting | Description |
+|---------|-------------|
+| `VAULT_MCP_CF_ACCESS_TEAM_DOMAIN` | Your Cloudflare team domain, e.g. `myteam.cloudflareaccess.com` (a bare `myteam` or a full URL is accepted). |
+| `VAULT_MCP_CF_ACCESS_AUD` | The Access application **audience (AUD)** tag from the Cloudflare dashboard. |
+
+Install the extra dependencies (PyJWT + cryptography):
+
+```bash
+pip install 'obsidian-web-mcp[cloudflare-access]'
+```
+
+#### Setting up Cloudflare (Managed OAuth)
+
+This walkthrough exposes the server through a Cloudflare Tunnel and puts Cloudflare
+Access + Managed OAuth in front of it, so an MCP client (e.g. Claude) authenticates
+against Cloudflare and Cloudflare injects the `Cf-Access-Jwt-Assertion` header the server
+verifies. It also shows where the two settings above come from.
+
+1. **Create the Access application.** In the Cloudflare dashboard go to
+   **Access → Access Controls → Create new application → Self-hosted and private →
+   Private destinations**. Set the **subdomain** (e.g. `my-vault-mcp`), the **domain** to
+   one you own (`domain-you-own.com`), and leave the **path** empty. This is the public
+   hostname (`my-vault-mcp.domain-you-own.com`) that your Cloudflare Tunnel forwards to the
+   server on loopback.
+2. **Enable an identity provider.** For the simplest setup, use the built-in
+   **`onetimepin`** provider, which emails a one-time code — enable it under
+   **Access → Identity providers**. (Any other SSO provider works too.)
+3. **Restrict who can get in.** Add an Access **policy** that allows only your own email
+   address (an *Allow* rule with an `Emails` / `Include` condition). This is your
+   per-identity, revocable gate — kill it in the dashboard to cut off access, no redeploy.
+4. **Enable Managed OAuth.** On the application's **Advanced settings** tab, turn on
+   **Managed OAuth** (it is opt-in for self-hosted applications). In **Allowed redirect
+   URIs** — the allowlist for dynamically-registered OAuth clients; entries must use
+   `https` and may end in `/*` — add the Claude client callbacks:
+   - `https://claude.ai/api/mcp/auth_callback`
+   - `https://claude.com/api/mcp/auth_callback`
+5. **Set sensible token lifetimes** (also on **Advanced settings**). Cloudflare uses two
+   durations: **Access token lifetime** (how long each token authenticates a request;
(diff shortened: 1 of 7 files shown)


jimprosser (changes requested):
Thanks for this. I ran the suite (415 passed with ripgrep, 414 + 1 skip without) and probed the verification through the real app with a live local JWKS server: alg=none, HS256 keyed with the RSA public key (hand-built), wrong key, unknown and missing kid, wrong aud and iss, expired, future nbf, redirecting JWKS, JWKS down after cache expiry. All 401, and the static bearer is refused in this mode. The design is sound and I want to merge it. A few things first, all small:

1. Bump the extra to `pyjwt>=2.14`. Redirect rejection and unknown-kid refresh rate-limiting landed in 2.14.0; cache-preserve-on-fetch-error in 2.13. At `>=2.8` the floor allows all three regressions. Note `mcp` already requires `pyjwt[crypto]>=2.10.1`, so the extra's job is the floor, not the install; adjust the README line accordingly.
2. `verify_access_token` runs a blocking urlopen (30 s timeout) inside `async dispatch`. Under uvicorn a concurrent `/health` took 2.7 s while a 3 s JWKS fetch was in flight. Wrap it in `await run_in_threadpool(...)`.
3. Half-config (exactly one of the two vars set) should exit at startup like the malformed-domain case, not warn and boot bearer+OAuth. Same "silently broken server" argument your docstring makes.
4. Service-token JWTs carry `sub: ""` with the identity in `common_name` (Cloudflare's application-token doc), so the audit principal is empty for them. Use `email or common_name or sub`. Also note in the README that in CF mode `client_id` is the plaintext identity and `token_id_hash` is its hash.
5. `test_hs256_token_signed_with_public_key_rejected` hits the `except: return` on PyJWT 2.14 and asserts nothing; `test_alg_none_token_rejected` has the same escape hatch. Build the forged tokens by hand and drop both. And `test_valid_cf_header_reaches_transport` asserts only `!= 401`; it actually returns 500 here. Assert a real MCP `initialize` succeeds.

Two more from a second reviewer, quick to fold in: a malformed team domain is echoed into the startup error verbatim, so a pasted `https://user:secret@myteam.cloudflareaccess.com/` lands the secret in the log (report the setting name and expected format, not the value), and PyJWKClient's `json.load` has no response-size cap.

Optional, your call: restrict the team domain to `.cloudflareaccess.com` (makes the "never SSRF" claim true by construction), a small `leeway`, and an `ALLOWED_EMAILS` pin so the origin enforces identity independently of the Access policy.

One thing to know: Cloudflare Access does not work with the claude.ai connector (its custom-header allowlist has no room for the Access headers, and Managed OAuth breaks RFC 8707 `resource`), so this mode is for other clients. Still worth having.

jimprosser:
Checking in on the 9/14 list: the pyjwt floor, the threadpool wrap, failing at startup on half-config, and the service-token principal. They're all small, and I'd still like to merge this once they're in. CI runs on PRs now (#90), so a rebase will show a green check.

PR #70, opened 2026-08-07
fix: stop leaking inotify watchers by making FrontmatterIndex a true process singleton

wtrchmr:
## Problem

`stateless_http=True` means `lifespan()` in `server.py` runs on **every**
MCP session, not once per process. It unconditionally called
`frontmatter_index.start()` on each session, and `FrontmatterIndex.start()`
unconditionally replaced `self._observer = Observer()` without stopping
the previous one — leaking one `watchdog.Observer` (and its inotify
watches) per session.

If `start()` itself raised (e.g. already at the inotify watch limit), the
paired `stop()` after `yield` was never reached either, compounding the
leak on every failed session.

In production this eventually exhausted `fs.inotify.max_user_watches`
(default 8192) after a few days of normal usage — every new MCP session
failed with `OSError: [Errno 28] inotify watch limit reached`, which
surfaced to clients (e.g. claude.ai's custom connector) as a hard
connection failure.

## Fix

- `frontmatter_index` is already a module-level singleton — this PR makes
  its lifecycle match that: `frontmatter_index.start()` now runs once in
  `main()` before the server starts accepting requests, instead of once
  per session. `lifespan()` is now just a thin `yield`.
- `FrontmatterIndex.start()` is now idempotent (no-op if already running)
  as defense in depth, in case anything else ever calls it more than once.

## Side effect

Removes the ~2 second full-vault reindex that previously happened on
every single stateless session, which was also a likely contributor to
intermittent timeouts under load.

Tested against a ~100-file vault: after the fix, "Starting vault MCP
server" logs exactly once per process lifetime (previously once per
session — 91 times in one day on the reporter's deployment), and the
process's inotify fd count stays flat across repeated sessions instead
of growing by ~197 watches each time.

### DIFF
--- a/src/obsidian_vault_mcp/frontmatter_index.py
+++ b/src/obsidian_vault_mcp/frontmatter_index.py
@@ -25,7 +25,16 @@ class FrontmatterIndex:
         self._pending_paths: set[str] = set()
 
     def start(self) -> None:
-        """Walk all .md files, parse frontmatter, and start watching for changes."""
+        """Walk all .md files, parse frontmatter, and start watching for changes.
+
+        Idempotent: a second call is a no-op if already running, so that an
+        accidental repeat call (e.g. from a future code path) can never
+        orphan a previous Observer/inotify watch instead of replacing it.
+        """
+        if self._observer is not None:
+            logger.warning("FrontmatterIndex.start() called while already running; ignoring")
+            return
+
         t0 = time.monotonic()
         count = 0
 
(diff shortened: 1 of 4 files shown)


jimprosser:
Good diagnosis, and it was real. Main has had this fix since June 7 (13e147f, #28): the index is built once in `main()`, `lifespan` is a no-op yield, `start()` is idempotent, and `atexit` stops the observer. Your branch predates it, which is why it conflicts in both files. The date normalization is also already on main via `serialization.py`, and the `allowed_hosts` entry for your hostname belongs in `VAULT_MCP_ALLOWED_HOSTS`, not the source. Rebasing your deployment onto current main should clear the inotify exhaustion. Closing.

PR #72, opened 2026-08-16
fix: claude.ai connector interop — OIDC discovery alias, root-level oauth endpoint aliases, longer keep-alive

hboylan:
Three small interop fixes (17 lines) that together let the claude.ai remote connector complete OAuth and stay connected. Observed against claude.ai as of Aug 2026, server exposed via Tailscale Funnel.

## Problems

1. **claude.ai probes `/.well-known/openid-configuration` and treats a non-2xx as fatal** instead of falling back to `/.well-known/oauth-authorization-server`, so the connector never reaches the authorize step.
2. **Clients that register via the RFC-default `/register` fallback assume root-level `/authorize` and `/token`** rather than reading endpoint URLs from AS metadata — those paths 404/401 here.
3. **uvicorn's 5s `timeout_keep_alive` closes idle sockets faster than claude.ai's connection pool (via the Funnel relay) notices**, so it reuses dead connections and surfaces intermittent 502s after a successful connect.

## Changes

- Serve the existing AS metadata handler at `/.well-known/openid-configuration` as an alias.
- Alias `/register`, `/authorize`, `/token` to the existing `/oauth/*` handlers and add them to the auth-exempt set.
- `timeout_keep_alive=75` (matches nginx's default).

No behavior change for clients already using the `/oauth/*` paths and standard discovery.

## Testing

Deployed on the fork behind Tailscale Funnel: claude.ai connector previously failed at auth; with this patch it completes the OAuth flow and lists tools/resources/prompts (verified in server logs). Local `/oauth/*` paths still work unchanged.

Related to the interop line of #19/#33; offered per #50.

https://claude.ai/code/session_01VfBnn64rraj4PUcwA7tUUU

### DIFF
--- a/src/obsidian_vault_mcp/auth.py
+++ b/src/obsidian_vault_mcp/auth.py
@@ -16,9 +16,13 @@ _AUTH_EXEMPT_PATHS = {
     "/health",
     "/.well-known/oauth-authorization-server",
     "/.well-known/oauth-protected-resource",
+    "/.well-known/openid-configuration",
     "/oauth/authorize",
     "/oauth/token",
     "/oauth/register",
+    "/register",
+    "/authorize",
+    "/token",
 }
 
 # (method, path) pairs exempt from auth. The MCP spec 2025-06-18 probe on / must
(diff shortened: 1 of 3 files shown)


hboylan:
Closing: retested against claude.ai today with stock main (cd88fdf) and the connector now completes OAuth cleanly — the client does RFC 9728 discovery and follows AS metadata to the /oauth/* endpoints, with no OIDC-discovery probe or root-level path fallback in the logs. The interop shims are no longer needed; claude.ai appears to have fixed this client-side.

PR #73, opened 2026-08-17
Harden OAuth with per-client tokens and server-side read-only policy

hawkxdev:
## Summary

- replace the shared bearer issued to dynamic OAuth clients with independent opaque access tokens backed by a versioned SQLite lifecycle store;
- authenticate every MCP request as an immutable principal and enforce the principal's stored capabilities in both tool discovery and dispatch;
- add operator-only client inventory/revoke/backup commands, credential-safe audit attribution, migration safeguards, and fail-closed extension registration/routing guards.

This is one security boundary rather than two independent features: issuing per-client tokens without enforcing their stored policy would still leave every dynamic client with the full MCP surface.

## Security behavior

- Dynamic clients authenticate with their own registered secret and receive unrelated `v1.*` access tokens with bounded expiry, canonical resource binding and durable revocation metadata.
- Authorization codes are PKCE-bound, client-bound, redirect-bound, resource-bound, single-use and consumed atomically.
- The `vault_readonly_v1` policy exposes exactly `vault_batch_read`, `vault_list`, `vault_read`, `vault_search` and `vault_search_frontmatter`.
- Mutation tools, resources, resource templates, prompts and protected non-MCP routes are denied before side effects for restricted principals.
- Master bearer and explicitly approved imported clients retain the existing full-access behavior.
- Unknown policies, invalid capabilities, expired/revoked tokens and wrong-resource credentials fail closed.
- Audit records contain stable hashed credential/client attribution but never raw bearer tokens or client secrets.
- Interactive consent identifies the escaped client name and exact registered redirect URI before credentials are submitted.
- Dynamic registration rejects empty, malformed, unsafe, or mixed-validity redirect metadata without persisting a client.
- Extensions receive a tool-only policy registrar and cannot shadow the MCP transport or auth-exempt routes; routes that cannot be safely inspected fail closed at startup.

## Compatibility and operations

- Existing `oauth_clients.json` data is imported only through the explicit `vault-mcp-oauth migrate` operator action and legacy allowlist; ordinary startup and metadata commands do not consume migration state.
- SQLite state, sidecars and backups enforce owner-only modes and reject unsafe ownership or symlink layouts before opening.
- Bearer authentication is read-only after state initialization and does not rewrite the configured static client on each request.
- The existing master bearer remains available for operator compatibility.
- The MCP dependency is pinned to the source-reviewed `mcp[cli]==1.28.1` contract; a clean install no longer resolves MCP 2.0.
- README and `.env.example` document state path, TTL, migration allowlist and local-only administration without exposing runtime credentials.

## Verification

Latest local verification on Python 3.14 after review remediation:

- full suite: `440 passed, 1 warning in 4.22s`;
- focused remediation regressions: `10 passed, 1 warning in 0.57s`;
- security mutation checks: `4 passed, 1 warning in 0.72s`;
- `python -m compileall -q src tests` passed;
- Ruff passed on every changed file under the repository's existing `E402` import-order baseline;
- gitleaks scanned all 4 PR commits and found no leaks.

The remaining warning is inherited from Pydantic settings through the MCP dependency; all tests pass.

## Review history

Independent local review found one restricted-principal bypass: an extension route could shadow the MCP transport path and inherit the middleware allowance intended for MCP dispatch. A regression test now proves both full and partial route matches are rejected before middleware attachment.

A follow-up review was verified against the implementation before changes were applied. The resulting remediation makes migration explicit, removes authentication-path writes, handles arbitrary Unicode credentials, exposes consent identity safely, rejects invalid redirect registration atomically, and turns unknown extension-route inspection into a startup failure. Lifecycle pruning, public token identifiers, and stricter symlink canonicalization remain non-blocking design cleanup rather than merge blockers.

## Checklist

- [x] One self-contained change, based directly on current main
- [x] New and abuse inputs are covered through real OAuth and MCP wiring
- [x] Auth routes and mount collisions are validated and fail closed
- [x] No new unchecked vault-path handling
- [x] No new request-controlled shell or outbound-request surface
- [x] No secrets, tokens or capability URLs in logs or committed evidence
- [x] New configuration is validated at startup and security defaults fail closed
- [x] No heavy or bridge dependency added
- [x] Existing atomic vault-write paths remain unchanged
- [x] README and `.env.example` document the new configuration and operations

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,24 @@
+# Required runtime paths and credentials
+VAULT_PATH=/absolute/path/to/ObsidianVault
+VAULT_MCP_TOKEN=replace-with-32-byte-random-hex
+VAULT_OAUTH_USERNAME=obsidian
+VAULT_OAUTH_PASSWORD=replace-with-an-interactive-login-secret
+
+# Pin the public OAuth resource in reverse-proxy deployments
+VAULT_MCP_PUBLIC_URL=https://vault-mcp.example.com
+VAULT_MCP_ALLOWED_HOSTS=vault-mcp.example.com
+
+# Persistent per-client OAuth lifecycle state. Keep this outside VAULT_PATH.
+VAULT_OAUTH_STATE_PATH=~/.local/share/vault-mcp/oauth_state.sqlite3
+VAULT_OAUTH_ACCESS_TOKEN_TTL_SECONDS=86400
+
+# One-way import from the legacy plaintext client registry. Leave the approval
+# list empty unless specific reviewed client IDs may temporarily retain
+# legacy_full policy during migration.
+OAUTH_CLIENTS_PATH=~/.local/share/vault-mcp/oauth_clients.json
+VAULT_OAUTH_APPROVED_LEGACY_CLIENT_IDS=
+
+# Optional static confidential client for client_credentials or browser flow.
+VAULT_OAUTH_CLIENT_ID=vault-mcp-client
+VAULT_OAUTH_CLIENT_SECRET=
+VAULT_OAUTH_REDIRECT_URIS=
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ All configuration is via environment variables:
 | Variable | Required | Default | Description |
 |----------|----------|---------|-------------|
 | `VAULT_PATH` | Yes | `~/Obsidian/MyVault` | Absolute path to your Obsidian vault directory |
-| `VAULT_MCP_TOKEN` | Yes | (none) | 256-bit bearer token validated on every MCP request |
+| `VAULT_MCP_TOKEN` | Yes | (none) | 256-bit master bearer. Protected requests also accept validated durable OAuth access tokens under their stored policy. |
 | `VAULT_OAUTH_PASSWORD` | **Yes** | (none) | Password for the interactive login at `/oauth/authorize`. **If unset, the server refuses to authorize any client (fail-closed).** |
 | `VAULT_OAUTH_USERNAME` | No | `obsidian` | Username for the interactive login |
 | `VAULT_MCP_HOST` | No | `127.0.0.1` | Bind address. Loopback by default; set `0.0.0.0` only for deliberate LAN exposure |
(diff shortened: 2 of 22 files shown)


thomastoye:
You might also be interested in #66, where I took the route of delegating auth to Cloudflare

jimprosser (changes requested):
Thanks for the work here; the test coverage is real and the token store is careful. I'm going to ask for changes rather than merge, because the direction is half right for this project.

The half I want: per-client access tokens replacing the master bearer we hand out today, with revocation. That closes a hole the code already documents as a follow-up.

The half I don't want: making every dynamically registered client read-only. This is a one-person server whose whole use is Claude reading and writing notes. The password at the consent screen is the authorization decision. There is also no command to grant full access afterwards, so the feature removes the product.

Three things need fixing before anything lands. Migration runs on ordinary startup and deletes `oauth_clients.json`; the PR text says it only runs via `vault-mcp-oauth migrate`, which does not exist. I reproduced it: one restart without the allowlist leaves both my connectors permanently `reauthorization_required`, and setting the allowlist afterwards does nothing. There are no refresh tokens, so every device re-enters the password every day. And `resource` plus `client_secret` are now hard-required at the token endpoint with no documentation, which will lock out any client that does not send them byte-identically.

Please split it. The `mcp<2` pin is already on main (#68); an exact `==1.28.1` pin would freeze us out of SDK security releases, so drop it. Then: per-client tokens with refresh, full access for all clients, non-destructive migration, no allowlist. The extension-route hardening (uninspectable routes fail closed; transport-path shadowing rejected) can be its own small PR now and I'd merge it. If you want to argue for read-only clients, open an issue and we can talk about an opt-in at consent time, default full.

jimprosser:
Checking in on the 9/14 review. The split still stands. First, per-client tokens with refresh, full access for every client, and a migration that leaves `oauth_clients.json` alone. Second, the extension-route hardening as its own small PR, which I'd merge quickly.

This branch can't land as it is, and main keeps moving under it (#93 moves the audit wrapper this PR edits). If I don't hear back by October 8 I'll close it. I want the per-client token work, so send the split PRs whenever they're ready.

PR #74, opened 2026-08-17
fix(deps): cap MCP below breaking v2

### LINKED ISSUE #71: Fresh installs broken: unbounded mcp dependency now resolves to the 2.0 SDK
Thanks for building this! ❤️ 

This is from Claude Code so please take with a grain of salt:

Fresh installs are currently broken because `pyproject.toml` declares `mcp[cli]>=1.9.0` with no upper bound, and the MCP Python SDK released 2.0.0, which removes the `mcp.server.fastmcp` module path that `server.py` imports. A clean `pip install` today resolves `mcp==2.0.0` and the server dies on startup:

```
Traceback (most recent call last):
  File "/usr/local/bin/vault-mcp", line 5, in <module>
    from obsidian_vault_mcp.server import main
  File "/usr/local/lib/python3.12/site-packages/obsidian_vault_mcp/server.py", line 12, in <module>
    from mcp.server.fastmcp import FastMCP
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

This affects both the v0.2.0 tag and current main (same unbounded constraint in both).

Pinning the SDK below 2 fixes it, so the quick patch is:

```toml
"mcp[cli]>=1.9.0,<2",
```

(with an eventual migration to the 2.x SDK as the longer-term fix).

I hit this while Dockerizing the server with Claude Code, and can confirm the `<2` pin builds and runs fine on my install (current main at cd88fdf).

One small related thing I noticed while poking at versions: the v0.2.0 tag is quite far behind main by now (no `/health` route, no `VAULT_MCP_PUBLIC_URL` pinning or `VAULT_MCP_FORWARDED_ALLOW_IPS`, and `forwarded_allow_ips="*"` is hardcoded in the tagged release), so anyone installing from the tag also misses most of the README's documented security features. A fresh tag once the pin lands would be lovely.

Happy to test a fix!

hawkxdev:
## Summary

- cap `mcp[cli]` below the incompatible 2.x major release;
- keep clean installs on the existing `mcp.server.fastmcp` contract.

Closes #71.

## Reproduction and verification

On a clean worktree from current `main`:

- before the change, `uv sync --extra dev` resolved `mcp==2.0.0` and pytest stopped during collection with 8 `ModuleNotFoundError: mcp.server.fastmcp` errors;
- after adding `<2`, the same command resolved `mcp==1.29.0` and `mcp.server.fastmcp` was importable;
- full suite: `373 passed, 2 warnings in 4.38s`;
- `python -m compileall -q src tests` passed;
- `git diff --check` passed.

The two warnings are inherited dependency deprecations from Pydantic settings and Starlette `TestClient`.

## Scope

One dependency constraint changes. No application, OAuth, vault, route, or deployment behavior changes.

## Checklist

- [x] One self-contained fix, based directly on current main
- [x] No new routes, request inputs, subprocesses, outbound requests, logs, or writes
- [x] No new dependency; the existing dependency is bounded to its compatible major line

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = "Secure remote MCP server for Obsidian vaults -- access your notes
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "mcp[cli]>=1.9.0",
+    "mcp[cli]>=1.9.0,<2",
     "python-frontmatter>=1.1.0",
     "ruamel.yaml>=0.18",
     "pydantic>=2.0",


jimprosser:
Closing as a duplicate of #68, which is the identical one-line change and was opened first. Your verification notes were useful, and #71 closed with it. Thanks.

PR #76, opened 2026-09-09
feat: vault_search matches note names/paths as well as contents

thomastoye:
## Problem

A note like `Trips/2026/NYC.md` is invisible to `vault_search("NYC")` unless its body happens to contain the query — both the ripgrep and Python backends only search file *contents*. Obsidian's own quick switcher matches paths, so this is a gap an LLM client hits constantly when a user refers to a note by name.

## Change

`vault_search` now runs a case-insensitive substring match of the query against vault-relative paths first, then the existing content search:

- Filename matches are returned first (a name hit is usually the strongest relevance signal), with `"match_type": "filename"`, `line_number: null`, and the relative path as `match_context`. Content matches follow, tagged `"match_type": "content"` and otherwise unchanged.
- The full relative path is matched, not just the stem — `"2026"` also finds `Trips/2026/NYC.md`.
- Filename matching honors `path_prefix`, `file_pattern`, `EXCLUDED_DIRS` (`.trash`, `.obsidian`, …), and the shared `max_results` budget, and behaves identically with or without ripgrep installed (it's a pure-Python pass either way — names only, so it's cheap even on large vaults).
- Both kinds of match get the existing frontmatter-excerpt enrichment.
- The duplicated walk/filter loop between the Python content fallback and the new filename pass is extracted into a shared `_iter_vault_files` generator.

Backward compatibility: existing content results only *gain* a `match_type` field. The one shape change is that filename results carry `line_number: null`.

Security-wise this adds no new surface: no new subprocess use (the filename pass never shells out), search stays under the resolved `path_prefix` via the existing `resolve_vault_path` guard, and excluded dotdirs are filtered as before.

## Tests

10 new tests in `tests/test_search_filenames.py` covering: name-only match, directory-component match, content tagging, ordering, `path_prefix` scoping, `file_pattern` filtering, excluded dirs, `max_results` truncation across both match kinds, frontmatter enrichment, and the no-ripgrep fallback.

There's no CI on this repo, so per CONTRIBUTING: full suite passes locally — `430 passed` on my integration branch and `383 passed` on this branch rebased on `main` (the delta is tests from my other open PRs). Note for anyone running the suite on a fresh venv: `pip install -e ".[dev]"` currently resolves `mcp` 2.x, which renamed `FastMCP`; install `'mcp<2'` to run against this codebase.


jimprosser (changes requested):
Thanks, this is a real gap and the shape is right: substring on the relative path, no new subprocess, name hits tagged and ordered first. Two things need to change before it lands, and one I'd like you to look at.

1. `_iter_vault_files` yields files ripgrep refuses. On the default backend, main never touched a symlinked file or a dot-directory note in `vault_search`; with this PR a symlink `vault/evil-link.md -> ~/outside/secret.md` matches by name and `_get_frontmatter_excerpt` reads its frontmatter through the link. Same for `.space/hidden-note.md`, which `resolve_vault_path` refuses to open. Verified through the registered tool. Please skip `is_symlink()` entries and any dot-prefixed path component in the walk (the same rule as `resolve_vault_path`), and drop the frontmatter read for filename-only hits altogether: a path is enough to locate the note, and the read is what makes this both a disclosure and an unbounded cost (a name-only match with `max_results=1` read an 8 MB file in full). Add a test that plants a symlink and a dot-directory note and asserts neither is returned.

2. Name hits eat the whole budget. With 30 notes under `Meetings/`, `vault_search("meeting")` on main returns the one body hit in `agenda.md`; on this branch it returns 20 filenames and the body search never runs. That drops results existing clients were getting, at the default cap, on any query that is also a folder or date token. Give the body search a guaranteed share (cap the name slice at `max_results // 2`, backfill afterwards), and turn `test_max_results_caps_combined_matches` into a test that a content hit survives when name hits exceed the cap. Dedupe by path while you're there; `banana.md` currently comes back twice.

3. The walk runs on every search even with rg installed, and `rglob` descends into `.git` and `.obsidian` before filtering. A 500-note vault with a 40k-object `.git` went from 17 ms to 416 ms per search here. Pruning excluded and dot directories during the walk (`os.scandir`, no per-entry stat) fixes it; `rg --files` when rg is present would give you parity for free. Not blocking, but please don't leave it as is.

Also add a positive `path_prefix` assertion (the current test only checks the empty case), a test through `server.vault_search` rather than only the helper, and update the docstring at `server.py:405`.

jimprosser:
Checking in on the 9/14 list: skipping symlinks and dot-directories in the walk (with no frontmatter read for name-only hits), a guaranteed share of `max_results` for content matches, and pruning excluded directories during the walk. I'd still like to merge this once those are in. CI runs on PRs now (#90), so a rebase will show you the result directly.

PR #77, opened 2026-09-13
Fix/pin mcp v1 lockfile

Darth-Ginger:


jimprosser:
Closing this one. The `mcp<2` pin landed via #68, and the file-permission fix landed via #67, which uses `fchmod` on the descriptor rather than `chmod` on the path and covers `write_bytes_atomic` too. Committing `uv.lock` is a separate policy question from either fix; if you want to make that case, open it as its own PR with the reasoning and nothing else in it. One change per PR, rebased on main, per CONTRIBUTING.

PR #80, opened 2026-09-16
Signed direct upload, rebuilt to the shape from the #64 review

mleitnercom:
Supersedes #64, rebuilt to the shape you described when closing it.

**Why, for me.** The feature is not hypothetical, though this code is new. My fork has run its own implementation of signed uploads since June, and I use it daily: 347 uploads on 64 days, median 0.56 MB, but 57 above 3 MB and 13 above 7.5 MB, the largest 29.5 MB. Those are scanned contracts, slide decks and screenshots that Claude or ChatGPT hand to the vault. `vault_write_binary` carries the file base64-encoded inside a tool argument, so the whole file goes through the model's context, and the clients give up long before 10 MB. With an upload URL the model only sees the URL, and the client sends the bytes.

**What this PR is and is not.** It is not the fork's code submitted upstream. It is a rebuild on current `main`, written against your review of #64. The fork runs a different implementation, with its own history and config. The two share one piece: the streaming design here also went into the fork today (v0.13.1), which is where the live numbers below come from.

**Your points from #64, one by one**

| Review point | Now |
|---|---|
| Route never worked (`config` not imported); test built its own app | Fixed. Every client-facing test requests the URL through the registered tool and posts through `build_app()`, bearer middleware included, no `Authorization` header |
| `VAULT_MCP_PATH=/upload/mcp/x` served the vault without a token | `/upload` is reserved in `_validate_mcp_path` |
| Extension route at `/upload/x/y` slipped past the guard | `build_app` rejects extension routes anywhere under `/upload`. The bearer exemption itself is only `POST` to exactly one id segment, so `/upload/a/b` and `GET /upload/<id>` need a token |
| Body buffered in full before id/signature check | Grant (id, signature, expiry, single use) is validated before any body byte is read |
| Stream to a tempfile under a hard cap | Body streams into a temp file in the staging dir, hashed on the way, cut off at the first chunk past the grant's cap. The route never calls `request.body()`. Multipart is refused |
| Stale sweep `rmtree`s any old directory | Only real directories named like a uuid4 upload id, never symlinks |
| Uploads bypass audit log and write listeners | Commit is audited as `vault_upload` with before/after snapshots and fires `created`/`updated`. Refused grants are logged only, so unauthenticated noise cannot grow the audit log |
| uvicorn access log records the signed URL | `serve()` passes a log config whose access handler redacts `signature=` |
| Cap is still `MAX_BINARY_SIZE`, nothing #61 cannot do | `VAULT_UPLOAD_MAX_BYTES` (default 100 MB), independent of `MAX_BINARY_SIZE`, since the body is a disk bound. A test uploads a file that `vault_write_binary` refuses |

Also: the grant is claimed atomically at commit (`mkdir`), so two requests on one URL cannot both write. A commit refused before writing releases the claim.

**Validation of this branch**

- `tests/test_signed_upload.py`, 23 tests. Run against #64's code with only the missing import added, 16 fail, each on the point it covers: MCP path accepted, extension route accepted, `POST /upload/a/b` served without a token, 64 of 64 body chunks pulled before refusing a bad signature and past a 1 KB cap, whole body buffered, no audit record, a foreign directory swept, the signature in a real uvicorn access log.
- Linux with ripgrep, tesseract and poppler installed: 446 passed, 0 skipped.

**The streaming design in production (my fork, not this branch)**

A 25 MB upload through a Cloudflare Tunnel took 6.2 s, SHA-256 identical, server RSS +2.2 MB. A second request on the same URL got 409.


jimprosser (changes requested):
Thanks for rebuilding this against the review instead of patching #64, and for the usage numbers; 57 uploads over 3 MB is a real reason for the route to exist. I ran the suite (447 passed, 0 skipped with ripgrep) and probed the route through `build_app()`. The hard parts hold, and this is the shape I asked for. A bad or expired grant is refused before `request.stream()` is touched. Swapping the target's parent directory for a symlink after the grant is issued gets a 400 and nothing lands outside the vault. A file that appears at the target after the grant gets a 409 with the original intact. `GET`, `PUT`, `/upload/<id>/`, `/upload/<id>/x`, `/upload` and `//upload/<id>` all get 401. The signature stays out of a real uvicorn access log. Three things block it, then a few small ones, and with those fixed I want it in.

1. It is on by default. With only `VAULT_MCP_TOKEN` set, the tool issues URLs signed with the bearer token and the tokenless route accepts them; upgrading turns on a new unauthenticated write path nobody asked for. CONTRIBUTING, under Config and defaults: "Anything new and security-relevant is off by default and opt-in." Register the tool and the route only when `VAULT_UPLOAD_URL_SECRET` is set, and drop the fallback to the bearer token. That also ends the question of whether one secret should be doing both jobs.

2. The new settings aren't validated at startup. `VAULT_UPLOAD_MAX_BYTES=100MB` silently becomes the default, `VAULT_UPLOAD_URL_TTL_SECONDS=-5` is accepted, and `VAULT_UPLOAD_STAGING_DIR` pointed inside the vault passes `validate_config()` although the README says it must be outside. The audit log has the same outside-the-vault rule and enforces it at startup; mirror that. Bad integers and non-positive values should exit with the variable's name, the way the heartbeat interval does.

3. The extension guard still has the #64 hole. Three probes cannot reserve a namespace: a literal `Route("/upload/export", h, methods=["POST"])` passes the guard and is served with neither a bearer token nor a signature, because the exemption regex matches it. `Route("/upload/{n:int}")` does the same. Literal `/upload/x/y` and `/upload/` also pass the guard; those still need a token, but the PR text says they are rejected, and `test_extension_route_in_the_upload_namespace_is_rejected` only tries parameterized routes. Add a plain prefix check on the route's `path` (equal to `/upload`, or starting with `/upload/`) next to the probes, and put the literal and typed cases in the test.

Smaller, all quick:

- `signature=%C3%A9` on a live grant returns 500 without a token: `hmac.compare_digest` raises `TypeError` on a non-ASCII `str`. Compare bytes.
- The log filter looks for the literal `signature=`. Starlette decodes query keys, so `%73ignature=<sig>` is accepted as the signature and logged in full. Because a wrong `Content-Type` refuses the request without burning the URL, the logged URL can still be live. Redact the whole query string for paths under `/upload/`; it's less code than the regex.
- `snapshot_path(target)` runs before the audit check, so every commit hashes the existing target (up to 100 MB) with auditing off, and an exception there leaves the grant claimed for good. Move it inside the `should_audit_operation` branch, or at least inside the claim's cleanup.
- Create the staging root and the per-grant directory with `mode=0o700`. They come out 0755 with `metadata.json` at 0644 under a normal umask.
- `tests/_live_server.py` inherits `os.environ`; see my note on #63, where the same file lives. Whichever lands second rebases onto the fix.

Two fixes landed on main today that you'll rebase over. #81 moves the middleware's exemption check from `request.url.path` to `request.scope["path"]`: today `POST /upload/a%3F/b` reads as `/upload/a` to the middleware while the router sees `/upload/a?/b`, so use the same variable for `is_signed_upload_request`. #82 puts `vault_write_binary` through `_run_audited` and `fire_write`, which it was missing, so the upload commit and the base64 path now report the same way.

I'm not asking for upload deadlines, a cap on concurrent receivers or outstanding grants, ownership proof in the stale sweep, or `create_dirs` in the signed string. Each needs a valid signed URL or a shell on the box to reach, and on a one-person server that is the owner.

mleitnercom:
All three blockers and the five smaller ones are done, on top of 34b1e67.

1. **Off by default.** `config.signed_upload_enabled()` is `bool(VAULT_UPLOAD_URL_SECRET)`, no fallback to the bearer token. With it unset the tool is not registered, `build_app` adds no route, and `is_signed_upload_request` returns False, so nothing is bearer-exempt. Proven in child processes started without the secret: `tools/list` does not contain `vault_request_upload_url` and `POST /upload/<id>` is 401.
2. **Validated at startup** when the feature is on, mirroring the audit log's rule: a non-integer or non-positive value, a maximum lifetime below the default lifetime, and a staging directory inside the vault each stop the server naming the variable. With the feature off nothing is validated, so a stale variable cannot block a server that does not use it.
3. **Namespace reserved for real.** The declared route path is checked (`/upload` or starting with `/upload/`) next to the probes, and the test now covers the literal, typed and deeper shapes. You were right that three probes could not do it: `Route("/upload/export")` passed and the exemption regex matched it.

Smaller ones: signature compared as bytes, so `signature=%C3%A9` is 403 instead of a 500. The access log redacts the whole query string under `/upload/` rather than matching the literal key, which `%73ignature=` slipped past. `snapshot_path` runs only when the record will be written, so a commit with auditing off no longer hashes the target and cannot leave the grant claimed on an error there. Staging root, grant directory and `metadata.json` are 0700/0600. `tests/_live_server.py` is the fixed version from #63; whichever lands second will carry it.

Against the previous revision 17 of the new tests fail, each on its own point. Server run with ripgrep, tesseract and poppler present: 484 passed, 0 skipped.

jimprosser (approved):
Re-checked against the 9/17 review. Off by default holds: with the secret unset the tool isn't listed and POST /upload/<id> is 401. Bad integers, non-positive values and a staging dir inside the vault all stop startup by name. The literal and typed routes under /upload are refused. A non-ASCII signature gets a 403, not a 500. The staging root is 0700. Suite is green merged onto current main (511 passed). Thanks for the thorough rebuild. Merging.

PR #88, opened 2026-09-19
fix: renaming a note through the MCP breaks every link to it

samlevan:
## The problem

Rename a note through this server and every link to it dies.

Say ten notes link to `[[Meeting Notes]]`. Ask Claude to rename it to `Standup Notes`. The file is renamed, the tool reports success, and those ten links now point at a note that does not exist. Nothing warns you. The notes look fine until you click one.

Rename that same note in the Obsidian app and all ten links follow it. So the same operation means two different things depending on which client you used, and the one that quietly damages the vault is the one an agent reaches for.

It is easy to miss because the damage is invisible at the moment it happens, and it compounds: every reorganisation from the phone or from Claude leaves more dead links behind. The server even ships a `broken_wikilinks` hygiene report, so today it can tell you about the links it just broke.

## The fix

`vault_move` now repoints those links, the way the app does. Rename `Meeting Notes` and the ten links read `[[Standup Notes]]` afterwards.

No new setting: nobody renames a note and wants the links to it left dangling, and the Obsidian app does not ask either. `vault_move` keeps its signature. No new env var, dependency or route.

## What it does, precisely

- Rewrites wikilinks and embeds (`[[t]]`, `[[t|alias]]`, `[[t#heading]]`, `[[t#^block]]`, `![[t]]`) with the alias, heading and block tail preserved, and markdown links whose target is a vault path, keeping a relative link relative.
- **A pure move touches nothing** that does not need it. Obsidian resolves `[[Shed]]` by name wherever the note sits, so moving a note between folders leaves those links alone; only a rename changes them.
- **An ambiguous name is never guessed.** If two notes are called `meeting.md`, Obsidian picks by proximity to the linking note, so rewriting `[[meeting]]` could silently repoint it at the other one. Those are left alone and reported back in the result instead.
- That ambiguity check runs against the vault **as it was before the move**, since the move has already landed by then and the survivor of a collision would otherwise look unique.
- **A failed rewrite never undoes the move.** It is logged and surfaced in the result; the move stands.
- Files that cannot be decoded or that exceed `MAX_CONTENT_SIZE` are skipped rather than rewritten, excluded and dot directories are never touched, writes go through `write_file_atomic`, and the moved note's own links are left as they are.

Result:

```json
{"source": "...", "destination": "...", "moved": true,
 "links": {"files_updated": 2, "links_updated": 7, "files": ["..."], "ambiguous": []}}
```

## Tests

`tests/test_link_updates.py`, 11 tests, all driving `vault_move` itself rather than the helper, so a disconnected wire fails. Includes the abuse cases: an undecodable file, an oversized file, `.trash`, a failed move and a path-traversal source, each asserting nothing was rewritten.

Full suite green locally: **467 passed, 6 skipped**.

https://claude.ai/code/session_011pXpSYsKuW8uuprnUhW3hS


jimprosser:
Thanks for this. It's a real gap, and yes to the design: on by default, no setting. Nobody renames a note and wants the links left dangling. Returning the list of changed files in the result is the right call. Two things block it, and there's one addition I'd like.

1. **The rewrite pass bypasses the read guards.** `_iter_markdown_files` reads every `.md` directly, so it skips the `st_nlink > 1` refusal from #79. I reproduced this on your branch. Hardlink an outside file into the vault as `notes.md` with a `[[target]]` in it. `vault_read` refuses it. Rename `target.md`, and the rewrite replaces the hardlink with an ordinary file that holds the outside content, which `vault_read` then returns in full. Skip symlinked files, and any file with `st_nlink > 1`, in the walk, the same rule the read paths use. Add a test for each.

2. **One failed write aborts the whole pass.** The `try` wraps the loop, not each file. A symlinked `.md` whose target is outside the vault makes `write_file_atomic` raise `Path resolves outside the vault root`. Every file after it goes unrewritten, and the vault ends up half-updated. The fix for (1) removes that trigger, but put the write in a per-file `try` anyway and report failures in the summary, so one bad file costs one file.

3. **A `dry_run` option on `vault_move`.** It should return the `links` summary, plus a diff for each file, without moving anything. This is the first tool call that can quietly edit dozens of files, so an agent (or I) should be able to see the blast radius before committing to it.

CI runs on PRs now (#90), so a rebase will show you the result directly.

